PR #35, opened 2025-10-07
feat(models): implement user and workflow models with timezone support and relationships

Spafic:


ashmod (dismissed):

services/api/.env.example
```diff
@@ -12,3 +12,6 @@ POSTGRES_DB=rune
 
 # Optional: Full database URL (overrides individual settings above)
 # DATABASE_URL=postgresql+asyncpg://postgres:postgres@localhost:5432/rune
+
+# Timezone for application timestamps (IANA timezone name). Example: Africa/Cairo
+APP_TIMEZONE=Africa/Cairo
```
unauthorised-401, line 17: Uses Docker container's system timezone - no need to set TZ variable

unauthorised-401 (changes requested):

ashmod (approved):

PR #79, opened 2025-10-15
fix(users): Implement case-insensitive email validation, UserRole enum with validation, and separate admin/self-service routes.

### LINKED ISSUE #65: [BE] User Management bug fix
# User Management Bug Fix

### 🔴 1. Case-Insensitive Email Validation

**Problem:**  
The current user creation process does not perform case-insensitive email validation. This allows duplicate accounts to be created with the same email address using different casing.

**Example:**  
- `user@example.com` and `User@Example.com` are treated as different emails

**Expected Behavior:**  
- Email uniqueness checks should be case-insensitive

---

### 🔴 2. Role Validation Missing

**Problem:**  
The system currently accepts any string value for the user role field without validation, leading to:
- Security vulnerabilities
- Data inconsistency
- Potential authorization bypass

**Expected Behavior:**  
- Implement strict validation against predefined allowed roles (Enums or database table for roles)
- Reject requests with invalid role values
- Return appropriate error messages

---

### 🔴 3. Route Architecture - Admin vs Self-Service

**Problem:**  
Current user routes conflate two distinct concerns:
- Users managing their own accounts (self-service operations)
- Admins managing other users' accounts (administrative operations)

This leads to unclear boundaries and potential security issues.

**Proposed Solution:**

#### Admin Routes (`/users/:id`)
- Purpose: Admin operations for managing other users
- Allowed updates: `name`, `email` only
- **Should NOT allow:** password updates (separate endpoint needed)

#### Self-Service Routes (`/me` or `/profile`)
- Purpose: Users managing their own accounts
- To be implemented with appropriate restrictions
- Can include password change functionality with proper verification

**Rationale:**  
While the current MVP doesn't include admin roles, implementing this separation now will:
- Prevent technical debt
- Improve code maintainability

Lucifer3224:
closes #65


services/api/src/users/router.py
unauthorised-401: Please remove the simple role check for now. It will be handled later by a dependency that verifies whether the user is an admin

services/api/src/users/router.py
```diff
@@ -0,0 +1,180 @@
...
+    GET /users/me
+    """
+    return ApiResponse(
+        success=True,
+        message="Profile retrieved successfully",
+        data=current_user,
+    )
+
+
+@router.put(
```
unauthorised-401: Please separate the password update logic from the general PUT /me route. There’s no need to implement the password change functionality now—just remove the option to update it here. Password changes will likely require additional verification or MFA in the future, and they’re not commonly done alongside profile updates like name or email

services/api/src/users/router.py
Lucifer3224: I understand your point and I've already added notes in the code for that. However, I think we should keep the simple role check for now because the route separation requires admin-only access to these endpoints.

services/api/src/users/service.py
```diff
@@ -46,7 +67,7 @@ async def get_user_by_email(self, email: str) -> User | None:
         Returns:
             User object if found, None otherwise
         """
-        statement = select(User).where(User.email == email)
+        statement = select(User).where(func.lower(User.email) == func.lower(email))
```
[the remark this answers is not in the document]
Lucifer3224: Okay but Just to clarify - can we use .lower() on [User.email] (a SQLAlchemy Column)? I thought .lower() only works on Python strings and we need func.lower() for Column objects to generate the SQL function. Or am I misunderstanding something?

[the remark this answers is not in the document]
unauthorised-401: What I meant in my comment was: you can use .lower() on the parameter email (which is a Python string), not on User.email (which is the Column).
So the code would be:
```python 
statement = select(User).where(User.email == email.lower()) 
```
Since all emails in the database are already stored in lowercase (enforced at creation by your code), we only need to lowercase the incoming email parameter. No need for func.lower() on the database column side at all

unauthorised-401 (dismissed):
Looks great! Thanks for your work on this. Approved @Lucifer3224

PR #105, opened 2025-10-22
feat(tests): Add tests for Auth

unauthorised-401:
* Add `freezegun==1.5.5` to the requirements to support time-based testing in the API service.
* Add Authentication testing


unauthorised-401:
Closes #106

PR #109, opened 2025-10-24
feat(templates): add CRUD operations and router for template management

Lucifer3224:


services/api/src/templates/service.py
```diff
@@ -0,0 +1,102 @@
...
+        result = await self.db.execute(stmt)
+        template = result.scalar_one_or_none()
+
+        if template:
+            template.usage_count += 1
+            await self.db.commit()
+            await self.db.refresh(template)
+
+    @staticmethod
+    def _extract_node_types(workflow_data: dict) -> tuple[Optional[str], Optional[str]]:
```
unauthorised-401: Could you please remove this functionality for now? Let’s start simple and only display the template name and description in the UI.

Also, since the nodes in the node list aren’t ordered — we rely on the edges to determine the direction — you could end up with any node as the “last” node. In branching workflows, there might even be multiple terminating nodes. To avoid that complexity for now, let’s skip this logic entirely

services/api/src/templates/service.py
unauthorised-401: Could you please replace the usage of execute() with exec()? This is the recommended approach in SQLModel, as execute() belongs to SQLAlchemy while exec() is the higher-level method provided by SQLModel.

Also, it would be better to use:

- .first() instead of scalar_one_or_none()

-  .all() instead of scalars().all()

These are more consistent with SQLModel’s query API and simplify the code a bit.

services/api/src/templates/service.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self, db: AsyncSession):
+        self.db = db
+
+    async def list_all_accessible_templates(
+        self, user_id: int
+    ) -> List[WorkflowTemplate]:
+        """Get all templates accessible to a user (public + their own)."""
+        stmt = select(WorkflowTemplate).where(
+            (WorkflowTemplate.is_public) | (WorkflowTemplate.created_by == user_id)
```
unauthorised-401: You can import and use or_ here ,that’s the recommended approach in SQLModel. It makes the intent clearer and follows the official documentation:
🔗 [SQLModel Docs – Where with multiple expressions using or_](https://sqlmodel.tiangolo.com/tutorial/where/#where-with-multiple-expressions-using-or)

services/api/src/templates/service.py
```diff
@@ -0,0 +1,102 @@
+from typing import List, Optional
```
unauthorised-401: You don’t need to import List from typing anymore — you can just use the built-in list type for annotations

services/api/src/templates/schemas.py
```diff
@@ -0,0 +1,43 @@
+from typing import Optional, Dict, Any
```
unauthorised-401: You don’t need to import Dict from typing anymore — you can just use the built-in dict type for annotations

services/api/src/templates/router.py
```diff
@@ -0,0 +1,101 @@
...
+async def delete_template(
+    template_id: int,
+    current_user: User = Depends(get_current_user),
+    service: TemplateService = Depends(get_template_service),
+) -> None:
+    """Delete a template."""
+    await service.delete_template(template_id, current_user.id)
+
+
+@router.post("/{template_id}/use", response_model=ApiResponse[dict])
```
unauthorised-401: The logic here looks good — but could you make the response directly return the workflow_data or define a proper response schema instead of using a plain dict?

Using a typed response model will give us type-safe auto-completion and make integration with the UI much smoother.

unauthorised-401 (changes requested):

Lucifer3224:
I've implemented the changes you requested. Kindly review and tell me if you'd like any additional adjustments. @unauthorised-401

services/api/src/templates/service.py
```diff
@@ -0,0 +1,85 @@
...
+    async def increment_usage_count(self, template_id: int) -> None:
+        """Increment the usage count for a template."""
+        stmt = select(WorkflowTemplate).where(WorkflowTemplate.id == template_id)
+        result = await self.db.exec(stmt)
+        template = result.scalar_one_or_none()
+
+        if template:
+            template.usage_count += 1
+            await self.db.commit()
+            await self.db.refresh(template)
```
Spafic: ### 🤔 Hmm, potential race condition in `increment_usage_count`
Hey! I noticed this function uses a read-modify-write pattern that might cause issues under concurrent requests.

**Current flow:**
1. Read `usage_count` from database (e.g., value is 10)
2. Increment in memory (`usage_count = 11`)
3. Write back to database

**The issue:** If two users clone the same template simultaneously, both might read `usage_count = 10`, both increment to `11`, and we end up with `11` instead of `12` in the database. The second update basically overwrites the first one! 

###  Recommended Fix

Use a **database-side atomic update** instead of the read-modify-write pattern.

**The idea:** Instead of fetching the template, modifying `usage_count` in Python, and saving it back, let the database do the increment directly with something like `SET usage_count = usage_count + 1`.

**Benefits:**
- The read and write happen atomically at the database level
- No race conditions — concurrent requests are handled correctly by the database
- Bonus: You can use `RETURNING` clause to get the updated template back if needed, all in one query!

Image

Spafic (changes requested):

services/api/src/templates/service.py
```diff
@@ -0,0 +1,85 @@
...
+    ) -> WorkflowTemplate:
+        """Create a new template."""
+        template = WorkflowTemplate(
+            name=template_data.name,
+            description=template_data.description,
+            category=template_data.category,
+            workflow_data=template_data.workflow_data,
+            is_public=template_data.is_public,
+            created_by=user_id,
+            usage_count=0,
```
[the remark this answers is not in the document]
Lucifer3224: It's not a critical issue, in fact it's not even an issue!

PR #112, opened 2025-10-26
feat(frontend): hook workflow UI into api backend

ashmod:
Workflow UI is now off mocks - the list, create modal, canvas save, and table actions now call the real `/workflows` endpoints, with dialogs and toasts handling feedback.

We still rely on placeholder metadata (triggerType, runs, lastRunAt, etc.) for the table view (and any future logging) until the backend surfaces those fields.
For now, “Run” only queues a stub message for now because the worker isn’t emitting `NodeExecutionMessages` yet.

Backend should handle worker queue integration, and exposing true workflow metadata.


ashmod:
PR is good to go, but let's wait until the workflow updates are handled in the BE so we can fully merge the feature.
cc @unauthorised-401

PR #119, opened 2025-10-28
feat(credentials): make credentials page and primary integration

SeifT101:
This pull request introduces major enhancements to the credentials management feature and expands the client SDK to support new API endpoints for both credentials and templates. The changes include a complete implementation of the credentials page with data fetching, creation, and deletion logic, as well as improvements to query parameter serialization and the addition of new endpoints in the generated client SDK.

The most important changes are:

**Credentials Management UI:**

* The `CreateCredentialsPage` (`page.tsx`) is now a fully functional, client-side page that fetches credentials from the backend, allows users to add new credentials, and provides a (placeholder) for deletion with optimistic UI updates and error handling. The previous placeholder UI has been removed.

**Client SDK Enhancements:**

* The generated client SDK (`sdk.gen.ts`) now includes new endpoints for credentials (`listCredentialsCredentialsGet`, `createCredentialCredentialsPost`, `listCredentialsDropdownCredentialsDropdownGet`) and templates (`listTemplatesTemplatesGet`, `createTemplateTemplatesPost`, `deleteTemplateTemplatesTemplateIdDelete`, `getTemplateTemplatesTemplateIdGet`, `useTemplateTemplatesTemplateIdUsePost`). These enable full CRUD and listing operations for both resources.

These changes lay the groundwork for robust credential and template management in the application, both in the UI and in API integrations.

### DIFF
--- a/apps/web/src/features/canvas/components/Inspector.tsx
+++ b/apps/web/src/features/canvas/components/Inspector.tsx
@@ -35,15 +35,13 @@ function renderInspectorForm(
   updateData: ReturnType<typeof useUpdateNodeData>,
   isExpanded: boolean,
 ) {
-  const commonProps = { node, updateData, isExpanded };
-
   switch (node.type) {
     case "http":
-      return <HttpInspector {...commonProps} />;
+      return <HttpInspector node={node} updateData={updateData} isExpanded={isExpanded} />;
     case "if":
-      return <IfInspector {...commonProps} />;
+      return <IfInspector node={node} updateData={updateData} isExpanded={isExpanded} />;
     case "smtp":
-      return <SmtpInspector {...commonProps} />;
+      return <SmtpInspector node={node} updateData={updateData} isExpanded={isExpanded} />;
     default:
       return null;
   }
--- a/apps/web/src/features/canvas/hooks/useUpdateNodeData.ts
+++ b/apps/web/src/features/canvas/hooks/useUpdateNodeData.ts
@@ -24,9 +24,9 @@ export function useUpdateNodeData(setNodes: SetNodes) {
           const nextData: NodeDataMap[T] = {
             ...n.data,
             ...updater(n.data as NodeDataMap[T]),
-          };
+          } as NodeDataMap[T];
 
-          return { ...n, data: nextData };
+          return { ...n, data: nextData } as CanvasNode;
         }),
       );
     },


apps/web/src/components/credentials/AddCredentialDialog.tsx
```diff
@@ -0,0 +1,325 @@
...
+      key: "host",
+      label: "SMTP Host",
+      type: "text",
+      placeholder: "smtp.gmail.com",
+      required: true,
+    },
+    {
+      key: "port",
+      label: "Port",
+      type: "text",
```
ashmod:
```suggestion
      type: "number",
```
Ports are numeric-only values.

apps/web/src/components/credentials/AddCredentialDialog.tsx
```diff
@@ -0,0 +1,325 @@
...
+}
+
+const CREDENTIAL_TYPES: { value: CredentialType; label: string }[] = [
+  { value: "api_key", label: "API Key" },
+  { value: "oauth2", label: "OAuth2" },
+  { value: "basic_auth", label: "Basic Auth" },
+  { value: "token", label: "Token" },
+  { value: "smtp", label: "SMTP" },
+  { value: "custom", label: "Custom" },
+];
```
ashmod: A bit of "re-inventing the wheel" is going on here. I believe credential types should not be duplicated in the frontend as this may lead to some issues and make maintenance more difficult.

Imagine the backend adds or removes a credential type, the frontend would still display it in the dropdown, user would be able to fill it but then would be met with a 422 from the API.

This is a tricky one but I think a more suitable approach would be some sort of utility file for credential types (e.g. `credentials/types.ts`) that maps the auto-generate CredentialType to human-friendly labels:
```typescript
 export const CREDENTIAL_TYPE_CONFIG: Record<
    CredentialType,
    {
      label: string;
      description?: string;
    }
  > = {
    api_key: {
      label: "API Key",
      description: "Simple API key authentication",
    },
    oauth2: {
      label: "OAuth2",
      description: "OAuth 2.0 authentication flow",
    },
    basic_auth: {
      label: "Basic Auth",
      description: "Username and password authentication",
    },
    token: {
      label: "Token",
      description: "Bearer token or access token",
    },
    smtp: {
      label: "SMTP",
      description: "Email server credentials",
    },
    custom: {
      label: "Custom",
      description: "Custom credential format",
    },
  } as const;

  /**
   * Get all credential types as an array for dropdowns.
   */
  export const getCredentialTypeOptions = () => {
    return (Object.keys(CREDENTIAL_TYPE_CONFIG) as CredentialType[]).map(
      (type) => ({
        value: type,
        label: CREDENTIAL_TYPE_CONFIG[type].label,
        description: CREDENTIAL_TYPE_CONFIG[type].description,
      })
    );
  };

  /**
   * Get label for a specific credential type.
   */
  export const getCredentialTypeLabel = (type: CredentialType): string => {
    return CREDENTIAL_TYPE_CONFIG[type].label;
  };
```
then this hardcoded array would just be:
```typescript
  const CREDENTIAL_TYPES = getCredentialTypeOptions();
```

And don't forget to update any dependents (e.g. Credentials Table)

This should generally mitigate the issue by failing on ts compiles if the schema ever changes in the backend.

apps/web/src/components/credentials/CredentialsTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+  if (h < 1) {
+    const m = Math.max(1, Math.floor(diff / (1000 * 60)));
+    return `${m}m ago`;
+  }
+  if (h < 24) {
+    return `${h}hr. ago`;
+  }
+  const d = Math.floor(h / 24);
+  return `${d}d ago`;
+}
```
ashmod: Another wheel re-inventing case, the problem with this custom function is that it doesn't effectively handle cases like pluralization (e.g. `1hr.` vs `2hrs.`), or abbreviations for other time units (months, days,..etc.) and any possible edge cases I may have missed.

I suggest replacing that with a library that could handle this for us, e.g. `date-fns`. I don't recall having this library installed in our packages, but it's tiny so I don't see any reason not to.

ashmod (commented):
Nice work! Small changes but otherwise solid job.

ashmod (approved):

PR #142, opened 2025-11-17
feat(user-management): implement password change

### LINKED ISSUE #140: [BE] Change Password
### Feature Summary

Create a password change feature that allows authenticated users to update their passwords securely by verifying their current password

### Problem / Use Case

Users need the ability to update their passwords for various reasons:

- Security best practices recommend periodic password changes

- Users may want to update weak passwords to stronger ones

- Compromised passwords need to be changed immediately

- Users sharing devices want to secure their accounts

### Proposed Solution

_No response_

### Additional Context

_No response_

### Pre-submission checklist

Lucifer3224:
This update improves how user passwords are created and managed.

Temporary passwords are now automatically generated using Python’s secrets module. This avoids weak or repetitive passwords that admins might create manually. It also prevents the security issue of admins knowing a user’s long-term password, the conventional and safer practice is to give users a temporary password and require them to set their own.

Users are required to change this temporary password on their first login. Until they do, access to sensitive actions is blocked. This ensures each user sets a password only they know.

After the password change, the session token is refreshed automatically, so the user can continue working without logging in again.

Admins can reset a user’s password at any time, generating a new temporary password and triggering the same flow.

**note:** Even if the frontend doesn’t have time to build a “set password” page, it’s not a blocker, we can simply redirect users to their existing profile page where they can change their password.

Overall, the goal is to improve security (stronger passwords, enforced change on first login) while keeping the experience simple.

closes #140

### DIFF
--- a/services/api/src/workflow/router.py
+++ b/services/api/src/workflow/router.py
@@ -10,7 +10,11 @@ from src.workflow.schemas import (
     WorkflowUpdateData,
 )
 from src.workflow.service import WorkflowService
-from src.core.dependencies import DatabaseDep, require_password_changed,get_current_user
+from src.core.dependencies import (
+    DatabaseDep,
+    require_password_changed,
+    get_current_user,
+)
 from src.workflow.dependencies import get_workflow_with_permission
 from src.workflow.permissions import require_workflow_permission
 from src.core.responses import ApiResponse


services/api/src/db/models.py
```diff
@@ -72,6 +72,9 @@ class User(TimestampModel, table=True):
     )
     is_active: bool = Field(default=True)
     last_login_at: Optional[datetime] = None
+    must_change_password: bool = Field(
+        description="Flag indicating user must change their password",
```
unauthorised-401: Can you add a default value here

unauthorised-401 (changes requested):
LGTM. I'll review the remaining files as soon as I can.

unauthorised-401 (changes requested):
I understand the reasoning for not forcing a password change, but since admins have the ability to perform highly destructive actions, I strongly prefer enforcing the mandatory password update. It’s important that they operate with a secure password of their own choosing rather than relying on a temporary one. @Lucifer3224

Lucifer3224:
Got it, that makes sense. Thanks for explaining the security side of it. I’ve updated the implementation so that newly created admins are now required to change their password before accessing any action. Let me know if you want anything adjusted.

PR #154, opened 2025-11-21
refactor(frontend): create page modifications

ShahdAshraff:


apps/web/src/components/layout/AppSidebar.tsx
```diff
@@ -139,6 +152,96 @@ function NavLink({
...
+          type="button"
+          className={cn(
+            "group relative flex h-12 w-full items-center rounded-xl border border-transparent text-sm font-medium text-muted-foreground motion-safe:transition-all motion-safe:duration-200 motion-safe:ease-out",
+            "focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-accent focus-visible:ring-offset-2 focus-visible:ring-offset-background",
+            "hover:border-border/70 hover:bg-muted/40 hover:text-foreground",
+            isExpanded ? "justify-start gap-3 px-3" : "justify-center gap-0 px-0",
+          )}
+        >
+          <Avatar className="h-8 w-8">
+            <AvatarImage src={user?.avatar_url ?? undefined} alt={user?.name ?? "User"} />
```
ashmod: We don't have an `avatar_url` configuration in our app. This needs to be removed.

apps/web/src/components/layout/AppSidebar.tsx
```diff
@@ -139,6 +152,96 @@ function NavLink({
...
+  const { user } = state;
+
+  const initials = user?.name
+    ? user.name
+        .split(" ")
+        .map((n) => n[0])
+        .join("")
+        .toUpperCase()
+        .slice(0, 2)
+    : "U";
```
ashmod:
```suggestion
  const initials = user?.name
    ? user.name
        .split(" ")
        .filter((n) => n.length > 0)
        .map((n) => n[0])
        .join("")
        .toUpperCase()
        .slice(0, 2)
    : "U";
```
If the name contains extra spaces (e.g., "John  Doe"), this will create empty strings in the array, and n[0] will be undefined. We need to filter out empty strings, as above.

PR #191, opened 2025-11-29
feat(frontend): user management page

ZeynatGhallab:
This includes: 1- Adding new button "Users" to the sidebar 
2- User Management page with invitation window to invite users and control their roles(Editor/Viewer)

### DIFF
--- a/apps/web/src/app/(app)/admin/users/page.tsx
+++ b/apps/web/src/app/(app)/admin/users/page.tsx
@@ -1,40 +1,25 @@
 "use client";
 
 import { useState, useEffect } from "react";
+import { useRouter } from "next/navigation";
 import { Card } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
 import { Avatar, AvatarFallback } from "@/components/ui/avatar";
-import { Users, CheckCircle, AlertCircle, Trash2, Edit2, Copy } from "lucide-react";
-import { Toaster, toast } from "sonner";
+import { Users, Trash2, Edit2, Copy } from "lucide-react";
+import { toast } from "@/components/ui/toast";
 import {
   getAllUsersUsersGet,
   createUserUsersPost,
   updateUserUsersUserIdPut,
   deleteUserUsersUserIdDelete,
 } from "@/client";
-import type { UserResponse, UserCreate, AdminUserUpdate } from "@/client/types.gen";
+import type { UserResponse, UserCreate, AdminUserUpdate, CreateUserResponse } from "@/client/types.gen";
 import { useAuth } from "@/lib/auth";
 
 export default function UsersPage() {
+  const router = useRouter();
   const { state } = useAuth();
   const currentUser = state.user;
-
-  // Admin guard: only admins can use this page
-  if (!currentUser) {
-    return <div className="p-8 text-sm text-muted-foreground">Loading user...</div>;
-  }
-  if (currentUser.role !== "admin") {
-    return (
-      <div className="p-8">
-        <Card className="p-6">
-          <h3 className="text-lg font-semibold">Access denied</h3>
-          <p className="text-sm text-muted-foreground mt-2">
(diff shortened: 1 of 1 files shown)


ashmod:
Nice work on the UI prototype, @ZeynatGhallab ! The layout and components are looking good 👍

A few things to wrap up before merging:
 
The main piece is wiring up the backend - I see you've got the TODO in place. The client types in @/client have been updated to match the API, so you should be good to go. You'll want to use `listUsers()` to fetch users, and `createUser({ name, email, role })` for the invite flow. 
Don't forget to display the `temporary_password` that comes back from the API so admins can share it with new users.

Also, the roles should be "user"/"admin" instead of "viewer"/"editor" to match what the backend expects.

For the 2FA toggle - let's remove it for now since we don't have that functionality in the backend yet. Better to add it later when and if it's actually wired up.

Lastly, the hardcoded values in the table ("Owner", "Today", "All projects") should come from the actual user data once you're pulling from the API.

Looking forward to seeing this come together!

ashmod:
Good changes! A few concerns:

1- Temporary password should be displayed in a separate window that the admin can close so they can copy the password at their own pace, instead of having it as a toast:
image
**Recommendation**: A modal with a copy-to-clipboard button would be ideal.

2- Admins should be able to remove and update users (those are available methods in the backend), currently there is no way of doing so in the user management page (admins are only allowed to just invite users, no configuration).
See "Admin-Specific Operation"s in `users/service.py` for a more detailed view.

Functionality is solid though! Once those couple issues are addressed we should be good to go!

apps/web/src/components/ui/switch.tsx
ashmod: This file is not used anywhere, it should be removed.

apps/web/src/app/(app)/admin/users/page.tsx
```diff
@@ -0,0 +1,260 @@
...
+  }, []);
+
+  // Invite user
+  const handleSendInvite = async () => {
+    if (!inviteEmail) return;
+
+    // Check if user already exists
+    const alreadyExists = users.some((u) => u.email.toLowerCase() === inviteEmail.toLowerCase());
+    if (alreadyExists) {
+      setNotification("User is already in the project");
```
ashmod:
```suggestion
      setNotification("User is already in the organization");
```

apps/web/src/components/layout/AppSidebar.tsx
```diff
@@ -244,6 +250,10 @@ function ProfileDropdown({ isExpanded }: { isExpanded: boolean }) {
 
 export function AppSidebar() {
   const pathname = usePathname();
+  const { state } = useAuth();                  
+  const user = state.user;
+  const isAdmin = user?.role === "admin";   // <-- detect admin
```
ashmod:
```suggestion
  const isAdmin = user?.role === "admin";
```

apps/web/src/app/(app)/admin/users/page.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  const [searchQuery, setSearchQuery] = useState("");
+  const [inviteOpen, setInviteOpen] = useState(false);
+  const [inviteEmail, setInviteEmail] = useState("");
+  const [inviteRole, setInviteRole] = useState<"user" | "admin">("user");
+
+  const [notification, setNotification] = useState("");
+  const [notificationType, setNotificationType] = useState<"success" | "error" | "info">("success");
+  const [tempPassword, setTempPassword] = useState("");
+  const [showNotification, setShowNotification] = useState(false);
```
ashmod: Our project includes **sonner** in `package.json`, but this file manually implements toast notifications with local state, custom timeouts, and CSS animations. This adds unnecessary complexity and bugs (e.g., the exit animation at L116 is unreachable because the component unmounts immediately when showNotification is false).

ashmod (changes requested):

PR #203, opened 2025-12-01
build(makefile): enhance api-install to create and use virtual environment

### LINKED ISSUE #202: [Improvement]: `make api-install` should create and use virtual environment
### Improvement Summary

The `make api-install` target currently installs dependencies without creating or using a virtual environment, which can pollute system Python and doesn't follow Python best practices. It should automatically create a venv if one doesn't exist, then install dependencies within it.

### Current Behavior

Currently, `make api-install` runs:
```makefile
api-install:
	@echo "Installing API dependencies..."
	cd services/api && pip install -r requirements.txt
```

**Problems:**
- Does not create a virtual environment if one doesn't exist
- Does not activate an existing virtual environment
- Installs packages to system Python (if venv not manually activated)
- Inconsistent with README instructions which show creating/activating venv first
- Can cause dependency conflicts with system Python packages
- Requires manual venv setup before running `make api-install`

### Proposed Improvement

The `api-install` target should automatically handle virtual environment setup:

**Step 1: Check for existing venv**
- Look for `venv` or `.venv` directory in `services/api/`

**Step 2: Create venv if missing**
- If no venv exists, create one using `python3 -m venv venv`
- Prefer `.venv` to match README convention, but support both

**Step 3: Activate venv and install**
- Activate the virtual environment
- Run `pip install -r requirements.txt` within the venv
- Provide clear feedback about which venv is being used

**Example behavior:**
```bash
$ make api-install
Setting up API dependencies...
Creating virtual environment...
Using virtual environment: services/api/venv
Installing dependencies...
✓ API dependencies installed
```

**Implementation approach:**
1. Check if `services/api/venv` or `services/api/.venv` exists
2. If neither exists, create `venv` directory
3. Activate the venv and install dependencies
4. Show clear messages to the user about what's happening

This ensures dependencies are always isolated in a virtual environment, following Python best practices and matching the README documentation.

### Benefits

- **Better developer experience**: New developers can run `make api-install` without manual venv setup
- **Dependency isolation**: Prevents polluting system Python with project dependencies
- **Consistency**: Matches the pattern shown in README and follows Python best practices
- **Fewer errors**: Reduces "module not found" errors from missing venv activation
- **Enables future improvements**: Once this is fixed, `api-server` can utilize `api-install` to ensure dependencies are ready

### Additional Context

- Related to issue #201 
- Once `api-install` creates venv properly, `api-server` can be simplified to use it
- Should support both `venv` and `.venv` for flexibility
- Other install targets (`web-install`, `worker-install`) don't have this issue as they use their own package managers

### Pre-submission checklist

DopeBiscuit:
- Automatically create venv if it doesn't exist
- Support both venv and .venv directory names
- Activate venv before installing dependencies
- Add api-install-no-env target for system Python installation with interactive confirmation
- Provide clear error messages if venv cannot be created
- Use POSIX-compatible . command instead of source for better compatibility

Fixes #202

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help install dev build clean docker-up docker-up-nginx docker-down docker-build docker-clean logs test lint format typecheck web-dev web-lint web-format api-dev api-lint api-format worker-dev worker-lint worker-format worker-test up nginx-up down nginx-down restart restart-nginx status
+.PHONY: help install dev build clean docker-up docker-up-nginx docker-down docker-build docker-clean logs test lint format typecheck web-dev web-lint web-format api-dev api-infra api-server api-install api-install-no-env api-lint api-format worker-dev worker-lint worker-format worker-test up nginx-up down nginx-down restart restart-nginx status
 
 # Default target
 help:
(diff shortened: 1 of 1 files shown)


ashmod:
Nice! Can you also update the README to reflect the refactors? It's using the old rules rn (e.g. `api-dev`).

PR #224, opened 2025-12-03
feat(frontend): scryb UI init

unauthorised-401:
created by @ashmod


ashmod:
This should be good to go now, we can add the markdown renderer later.

PR #234, opened 2025-12-05
fix(auth): display error message on invalid login instead of refreshing

EhDaYaGhaly:
Refactors the sign-in form to use React 19’s useActionState hook instead of react-hook-form, fixing the issue where the form would refresh the page on invalid credentials instead of displaying an error message.

Changes

* Replace react-hook-form with React 19’s useActionState hook
* Use native form action prop instead of onSubmit with handleSubmit
* Manage form state (error, pending) via useActionState
* Simplify loading and error state handling

### DIFF
--- a/apps/web/src/components/auth/SignInForm.tsx
+++ b/apps/web/src/components/auth/SignInForm.tsx
@@ -1,52 +1,80 @@
 "use client";
 
+import { useActionState } from "react";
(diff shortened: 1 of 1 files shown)


ashmod:
Thanks for this! Our auth is currently a work in progress so this change is on our plans, but not at this time. I'm closing this for now and will take a look on our next auth sprint.

PR #237, opened 2025-12-07
feat(credentials): add support for other authentication types

unauthorised-401:
This pull request introduces support for a new "header" credential type throughout the system, enabling users to authenticate HTTP requests using custom headers. The main changes add the new type to credential definitions, update UI components to allow selection and input of header credentials, and ensure backend services correctly handle and inject these credentials during execution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,8 +120,9 @@ worker-install:
 # Development targets
 # ======================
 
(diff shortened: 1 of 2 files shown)


DopeBiscuit (approved):
LGTM.

PR #242, opened 2025-12-07
feat(ui): add delete credentials

SeifT101:
Adds delete functionality for user created credentials on the frontend.

(Note: would need to revisit after role based access has been integrated)


ashmod:
LGTM. Thanks!

PR #245, opened 2025-12-08
feat(frontend): wire credential sharing functionality

Lucifer3224:
Some files are just reformatted with Ruff


Lucifer3224:
Really sorry for these mistakes! I guess I didn't have enough coffee so my brain had a race condition while writing this code.
I've implemented the fixes you mentioned, so let me know if there's anything else that needs fixing!

PR #257, opened 2025-12-12
feat: add scheduler trigger

### LINKED ISSUE #250: [Feature]: Scheduler Implementation
## Task Description
Implement scheduling system to support time-based workflow triggers alongside existing manual triggers.

---

### 1. Background Scheduler Service
- Create background process that polls database at configurable intervals (env variable 1)
- Search for tasks scheduled to run within lookahead window (env variable 2)
- Execute workflows at scheduled times

### 2. Smart Workflow APIs
- Detect trigger type (manual vs scheduled) in requests
- Route to appropriate service based on trigger type
- Create workflow triggers table for scheduled configurations
- Manual triggers: no configuration persistence needed
- Scheduled triggers: persist schedule patterns and timing data

### 3. Docker Updates
- Add scheduler service to docker-compose alongside API service
- Configure environment variables for polling intervals
- Add health checks for both services
- Ensure both processes share database and business logic

### 4. Frontend Integration
- Add trigger configuration UI in workflow editor
- Implement schedule configuration form with pattern input
- Add Active/Inactive toggle with visual status indicators
- Display next run time and last execution status
- Implement warning modals:
  - Activation warning: show schedule details and next run
  - Deactivation confirmation: explain impact
- Update workflow list to show trigger status and schedule info

### Problem / Use Case

We need to schedule a workflow to start on 1/1/2026 at 16:40:03 and repeat every minute. The workflow should be configurable to run for specific periods, such as one week, with the ability to activate and deactivate it as needed.

### Proposed Solution

_No response_

### Additional Context

_No response_

### Pre-submission checklist

Spafic:
- closes #250 

I have applied the specifications from board task #250 to the API service, frontend, and CI/CD pipeline.

### DIFF
--- a/apps/web/src/features/canvas/components/inspectors/ScheduledInspector.tsx
+++ b/apps/web/src/features/canvas/components/inspectors/ScheduledInspector.tsx
@@ -24,6 +24,16 @@ function intervalToSeconds(days: number, hours: number, minutes: number, seconds
   return days * 86400 + hours * 3600 + minutes * 60 + seconds;
 }
 
+function toLocalDateTimeInputValue(value: string) {
+  const date = new Date(value);
+  if (Number.isNaN(date.getTime())) {
+    return "";
+  }
+
+  const offsetDate = new Date(date.getTime() - date.getTimezoneOffset() * 60000);
+  return offsetDate.toISOString().slice(0, 16);
+}
+
 export function ScheduledInspector({
   node,
   updateData,
@@ -115,7 +125,7 @@ export function ScheduledInspector({
           className="w-full rounded-[calc(var(--radius)-0.25rem)] border border-input bg-muted/30 px-2 py-1 text-sm"
           value={
             node.data.start_at
-              ? new Date(node.data.start_at).toISOString().slice(0, 16)
+              ? toLocalDateTimeInputValue(node.data.start_at)
               : ""
           }
           onInput={(e) => handleStartTimeChange(e.currentTarget.value)}
--- a/apps/web/src/components/workflows/WorkflowsTable.tsx
+++ b/apps/web/src/components/workflows/WorkflowsTable.tsx
@@ -56,6 +56,7 @@ import {
 import { Button } from "@/components/ui/button";
 import { Label } from "@/components/ui/label";
 import {
+  canChangeWorkflowStatus,
   canExecuteWorkflow,
   canDeleteWorkflow,
   canShareWorkflow,
@@ -434,18 +435,30 @@ export function WorkflowsTable() {
               </TableCell>
               <TableCell className="text-muted-foreground">
                 {w.triggerType === "scheduled" ? (
-                  <button
-                    type="button"
-                    onClick={() => handleToggleSchedule(w)}
-                    disabled={isRowPending(w.id)}
-                    className={`rounded px-2 py-0.5 text-xs font-medium transition-colors ${
-                      w.scheduleActive
-                        ? "bg-green-900/30 text-green-300 hover:bg-green-900/50"
-                        : "bg-zinc-800 text-zinc-400 hover:bg-zinc-700"
-                    }`}
-                  >
-                    {w.scheduleActive ? "Active" : "Paused"}
-                  </button>
+                  canChangeWorkflowStatus(w.role, isAdmin) ? (
+                    <button
+                      type="button"
+                      onClick={() => handleToggleSchedule(w)}
+                      disabled={isRowPending(w.id)}
+                      className={`rounded px-2 py-0.5 text-xs font-medium transition-colors ${
+                        w.scheduleActive
+                          ? "bg-green-900/30 text-green-300 hover:bg-green-900/50"
+                          : "bg-zinc-800 text-zinc-400 hover:bg-zinc-700"
+                      }`}
+                    >
+                      {w.scheduleActive ? "Active" : "Paused"}
+                    </button>
+                  ) : (
+                    <span
+                      className={`rounded px-2 py-0.5 text-xs font-medium ${
+                        w.scheduleActive
+                          ? "bg-green-900/30 text-green-300"
+                          : "bg-zinc-800 text-zinc-400"
+                      }`}
+                    >
+                      {w.scheduleActive ? "Active" : "Paused"}
+                    </span>
+                  )
                 ) : (
                   "\u2014"
                 )}
@@ -548,7 +561,7 @@ export function WorkflowsTable() {
                         </DropdownMenuItem>
                       )}
                       {w.triggerType === "scheduled" &&
-                        canExecuteWorkflow(w.role, isAdmin) && (
+                        canChangeWorkflowStatus(w.role, isAdmin) && (
                           <DropdownMenuItem
                             onSelect={() => handleToggleSchedule(w)}
                             disabled={isRowPending(w.id)}


services/api/src/db/models.py
```diff
@@ -270,3 +300,53 @@ class WorkflowCredential(TimestampModel, table=True):
...
+        description="When this schedule should start (first execution time)",
+    )
+    next_run_at: datetime = Field(
+        description="When the next execution is scheduled",
+    )
+    last_run_at: Optional[datetime] = Field(
+        default=None,
+        description="When the last execution occurred",
+    )
+    run_count: int = Field(
```
unauthorised-401: I suggest removing run_count, failure_count, and last_error from this model. These are computed values that should be derived directly from the executions table/relation. Storing them here creates a 'dual source of truth' and risks the counters becoming out of sync with the actual logs

services/api/Dockerfile.scheduler
```diff
@@ -0,0 +1,18 @@
+# Scheduler Service Dockerfile
+FROM python:3.11-slim
+
+WORKDIR /app
+
+# Install dependencies
+COPY requirements.txt .
```
unauthorised-401: t looks like we are installing the global requirements.txt here, which includes dependencies for the entire project. Since the Scheduler only needs a small subset of these, this unnecessarily increases the Docker image size

unauthorised-401 (changes requested):

services/api/src/db/models.py
```diff
@@ -27,12 +27,23 @@ class CredentialType(str, Enum):
...
     CUSTOM = "custom"
     SMTP = "smtp"
 
 
+class TriggerType(str, Enum):
+    """Workflow automatic trigger type enumeration.
+
+    Note: ALL workflows can be manually run via API/UI.
+    This enum only tracks AUTOMATIC trigger types.
+    If trigger_type is NULL, workflow has no automatic triggers.
```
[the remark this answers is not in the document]
Spafic: It's just a design choice, we can make it manual.

services/api/Dockerfile.scheduler
Spafic: Resolved in New Arch.

Spafic:
> I have started the review but need to step away. I will resume and complete the review of the remaining files as soon as possible

Tyt no rush, Thank you.

services/scheduler/executor.py
```diff
@@ -0,0 +1,194 @@
...
+                    f"after error: {update_error}"
+                )
+
+    async def _update_schedule_after_execution(
+        self,
+        conn: asyncpg.Connection,
+        schedule_id: int,
+        interval_seconds: int,
+        success: bool,
+        error_message: Optional[str] = None,
```
unauthorised-401: Parameter is never used in the function

services/scheduler/main.py
```diff
@@ -0,0 +1,208 @@
...
+
+            except Exception as e:
+                logger.error(f"Error in polling loop: {e}")
+                logger.error(traceback.format_exc())
+
+            await asyncio.sleep(Config.POLL_INTERVAL)
+
+    async def _healthcheck_loop(self):
+        """Periodic healthcheck to monitor service health."""
+        await asyncio.sleep(Config.HEALTHCHECK_INTERVAL)
```
unauthorised-401: Config.HEALTHCHECK_INTERVAL is referenced but not defined in config.py

services/scheduler/main.py
```diff
@@ -0,0 +1,208 @@
...
+                    logger.error(
+                        f"HEALTHCHECK FAILED | "
+                        f"DB: {'OK' if db_healthy else 'FAIL'} | "
+                        f"MQ: {'OK' if mq_healthy else 'FAIL'}"
+                    )
+
+            except Exception as e:
+                logger.error(f"Healthcheck error: {e}")
+
+            await asyncio.sleep(Config.HEALTHCHECK_INTERVAL)
```
unauthorised-401: Config.HEALTHCHECK_INTERVAL is referenced but not defined in config.py

services/scheduler/credentials.py
```diff
@@ -0,0 +1,109 @@
...
+                    query = """
+                        SELECT id, name, credential_type, credential_data
+                        FROM workflow_credentials
+                        WHERE id = $1
+                    """
+                    credential = await db_conn.fetchrow(query, credential_id)
+
+                    if not credential:
+                        logger.error(f"Credential with ID {credential_id} not found")
+                        continue
```
unauthorised-401: Silent failure on missing credentials, using continue may leave workflow in invalid state

services/scheduler/database.py
```diff
@@ -0,0 +1,67 @@
...
+    async def connect(self) -> asyncpg.Pool:
+        """Establish database connection with retry logic."""
+        for attempt in range(1, self.max_retries + 1):
+            try:
+                logger.info(
+                    f"Connecting to database (attempt {attempt}/{self.max_retries})..."
+                )
+                self.pool = await asyncpg.create_pool(
+                    self.dsn,
+                    min_size=2,
```
unauthorised-401: Reduce min_size from 2 to 1 (50% idle connection reduction)

services/scheduler/database.py
```diff
@@ -0,0 +1,67 @@
...
+        self.pool: Optional[asyncpg.Pool] = None
+
+    async def connect(self) -> asyncpg.Pool:
+        """Establish database connection with retry logic."""
+        for attempt in range(1, self.max_retries + 1):
+            try:
+                logger.info(
+                    f"Connecting to database (attempt {attempt}/{self.max_retries})..."
+                )
+                self.pool = await asyncpg.create_pool(
```
unauthorised-401: Make pool size configurable using environment variables

services/api/src/workflow/router.py
```diff
@@ -75,36 +98,39 @@ async def create_workflow(
...
+    """
     wf = await service.create(
         user_id=current_user.id,
         name=payload.name,
         description=payload.description,
         workflow_data=payload.workflow_data,
     )
 
+    # If workflow has schedule configuration, create the schedule automatically
+    schedule_config = service._extract_schedule_config(payload.workflow_data)
```
unauthorised-401: router directly calls service._extract_schedule_config() - a private method (prefixed with _). This breaks encapsulation.

apps/web/src/lib/workflows.ts
```diff
@@ -52,11 +59,26 @@ export const defaultWorkflowSummary: WorkflowSummary = {
...
   item: WorkflowListItem,
 ): WorkflowSummary {
+  const triggerType = item.trigger_type === "scheduled" 
+    ? "Scheduled" 
+    : item.trigger_type === "webhook" 
+    ? "Webhook" 
+    : "Manual";
+  
+  // Determine status based on is_active flag (defaults to inactive if no schedule)
+  const status: "active" | "inactive" = item.schedule?.is_active ? "active" : "inactive";
```
unauthorised-401:
```suggestion
  const status = item.schedule?.is_active ? "active" : "inactive";
```
TypeScript will automatically infer the union type "active" | "inactive" here, so we can remove the explicit type annotation to keep it cleaner.

apps/web/src/features/canvas/components/inspectors/ScheduledInspector.tsx
```diff
@@ -0,0 +1,157 @@
...
+            node.data.start_at
+              ? node.data.start_at.slice(0, 16)
+              : ""
+          }
+          onChange={(e) =>
+            updateScheduledData((d) => ({
+              ...d,
+              start_at: e.target.value || undefined,
+            }))
+          }
```
ashmod: onChange is not ideal here, because `datetime-local` onChange only fires on commit, if the user closes the inspector before commit (e.g. they add a date, but go off-focus then go back to fill time, the values are reset). This doesn't match the behavior of other inspectors where they persist values per-keystroke since they have onChange is simpler there, it updates numeric values.

This can be fixed in many ways, from creating our own date-time component (most complex probably), or capturing `onInput` or buffering local state and committing on blur so the behavior would match the rest of the inspectors as much as possible.

apps/web/src/features/canvas/components/inspectors/ScheduledInspector.tsx
```diff
@@ -0,0 +1,157 @@
...
+        </label>
+        <input
+          type="datetime-local"
+          required
+          className="w-full rounded-[calc(var(--radius)-0.25rem)] border border-input bg-muted/30 px-2 py-1 text-sm"
+          value={
+            node.data.start_at
+              ? node.data.start_at.slice(0, 16)
+              : ""
+          }
```
ashmod: A tiny UX nit here, currently users must type a full datetime-local value, which is tedious and error prone. We could add quick-select start-time shorcuts (e.g., "Now", "+5m", "+15m", "+30m") to reduce manual datetime entry.

services/api/src/workflow/router.py
```diff
@@ -128,13 +154,53 @@ async def update_workflow_data(
...
+                is_active=schedule_config.get("is_active", existing_schedule.is_active),
+            )
+        else:
+            # Create new schedule
+            await schedule_trigger_service.create_schedule(
+                workflow_id=wf.id,
+                interval_seconds=schedule_config["interval_seconds"],
+                start_at=schedule_config.get("start_at"),
+                is_active=schedule_config.get("is_active", True),
+            )
```
ashmod: This could be the culprit in the bug where first save fails but second one passes.. there is schedule management logic here, but it's already created/updated in `service.py`.. So probably what's going on is: [`WorkflowService.update_workflow_data`](https://github.com/rune-org/rune/blob/3cf354754919871f9a9134574568ea132b1529e7/services/api/src/workflow/service.py#L159-L160)already creates/updates/deletes `ScheduledWorkflow`, but the router here re-computes `schedule_config` using `existing_schedule` captured before the update and calls `create_schedule` *again*, which raises the bad request. I think we should keep schedule management solely in the service and remove it entirely from router.

Spafic: slep

PR #258, opened 2025-12-12
feat(frontend): role based access

ShahdAshraff:


apps/web/src/lib/state.tsx
```diff
@@ -107,11 +109,39 @@ export function AppStateProvider({ children }: { children: ReactNode }) {
...
+                return p.user_id === userData.id || String(p.user_id) === String(userData.id);
+              }
+            );
+            
+            if (currentUserPermission) {
+              console.log(`✓ Found role for workflow ${item.id}: ${currentUserPermission.role}`);
+              return { ...listItemToWorkflowSummary(item), role: currentUserPermission.role };
+            }
+            
+            console.warn(`⚠ User not found in permissions for workflow ${item.id}, defaulting to viewer`);
```
ashmod:
```suggestion
```

apps/web/src/lib/state.tsx
```diff
@@ -107,11 +109,39 @@ export function AppStateProvider({ children }: { children: ReactNode }) {
...
+            return { ...listItemToWorkflowSummary(item), role: "viewer" as WorkflowRole };
+          } catch (error) {
+            console.error(`❌ Failed to fetch permissions for workflow ${item.id}:`, error);
```
ashmod:
```suggestion
```

apps/web/src/lib/state.tsx
```diff
@@ -121,14 +151,48 @@ export function AppStateProvider({ children }: { children: ReactNode }) {
...
+      const workflowsWithRoles = await Promise.all(
+        items.map(async (item) => {
+          try {
+            const permissions = await listWorkflowPermissions(String(item.id));
+            console.log(`Workflow ${item.id} permissions:`, permissions);
+            console.log(`Current user ID: ${userData.id}`);
+            
+            const currentUserPermission = permissions.find(
+              (p) => {
+                console.log(`Comparing permission user_id ${p.user_id} (${typeof p.user_id}) with current user ${userData.id} (${typeof userData.id})`);
```
ashmod:
```suggestion
```

apps/web/src/lib/workflows.ts
```diff
@@ -47,16 +53,22 @@ export const defaultWorkflowSummary: WorkflowSummary = {
...
   runs: 0,
+  role: "owner", // Default to owner until backend sends actual role
 };
 
 export function listItemToWorkflowSummary(
   item: WorkflowListItem,
 ): WorkflowSummary {
+  // Backend doesn't yet send role in WorkflowListItem response
+  // When it does, use: (item as any).role || "owner"
+  const role: WorkflowRole = (item as any).role || "owner";
```
ashmod: This is type-unsafe and concerning logic.. I don't think this should be how that case should be handled.
cc @Spafic

apps/web/src/components/workflows/ShareWorkflowDialog.tsx
Spafic:
```suggestion
          <DialogTitle>Share &quot;{workflowName}&quot;</DialogTitle>
```

Spafic (dismissed):
@ShahdAshraff Minor fixes, otherwise LGTM!

apps/web/src/lib/state.tsx
```diff
@@ -107,11 +109,33 @@ export function AppStateProvider({ children }: { children: ReactNode }) {
...
+            if (currentUserPermission) {
+              return { ...listItemToWorkflowSummary(item), role: currentUserPermission.role };
+            }
+            
+            return { ...listItemToWorkflowSummary(item), role: "viewer" as WorkflowRole };
+          } catch (error) {
+            return { ...listItemToWorkflowSummary(item), role: "viewer" as WorkflowRole };
+          }
+        })
+      );
```
ashmod: This block (and the duplicate one below in `refreshWorkflows()`) makes a separate `listWorkflowPermissions()` API call for every workflow in the list. If someone has 30+ workflows, that's 30+ extra HTTP requests on every page load.

This is already solved in #373 which added the role field directly to the `WorkflowListItem` response from `GET /workflows`. You're already mapping it in `listItemToWorkflowSummary()` over in workflows.ts:

 ```ts
role: item.role,
```

So all of this Promise.all + listWorkflowPermissions code can be removed, and you can go back to the simpler version:
```ts
  dispatch({
    type: "setWorkflows",
    workflows: workflowItems.map((item) => listItemToWorkflowSummary(item)),
  });
```

Same applies to `refreshWorkflows()` below, it has the same pattern and can be simplified the same way. This also removes the need to fetch the user profile again in `refreshWorkflows()`.

services/api/openapi.yaml
```diff
@@ -673,23 +673,21 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPValidationError'
-  /users/directory:
+  /users/sharing:
```
ashmod: The spec now declares `GET /users/sharing`, but the backend router still exposes `GET /users/directory`.

Because `openapi.yaml` is generated from backend routes, running `python -m scripts.generate_openapi` then client regen (i.e. what the ci action does) will revert this path back to `/users/directory`..

Can we align these before merge (either keep `/users/directory` in OpenAPI, or update backend route and regenerate both OpenAPI + client accordingly)?

Also, we should avoid manual edits to `openapi.yaml`; changes should come from backend route/schema updates and then regeneration.

unauthorised-401: Just to clarify, /users/directory is the new intended route following the backend refactoring, so we should stick with that one. If you just regenerate the OpenAPI spec and the client from the backend, it will align everything and we should be good to go @ashmod

PR #273, opened 2025-12-20
refactor: move user sharing search from client to server

### LINKED ISSUE #267: [BE] Server Side search for users and credentials
One concern: The sharing dialog (`ShareWorkflowDialog`) currently fetches the entire user list via `listUsersForSharingUsersSharingGet` (wrapping GET/users/sharing). This will severely impact performance as we scale (both in network payload and rendering the Select component). I think we need to prioritize refactoring this to a server-side search/autocomplete endpoint in the backend. I also believe this problem affects credential sharing since they rely on the same endpoint.

_Originally posted by @ashmod in https://github.com/rune-org/rune/issues/258#issuecomment-3651642720_

Lucifer3224:
closes #267


ashmod:
to be revisited after ongoing refactors

ashmod:
@Lucifer3224 we should close this now, right?

Lucifer3224:
> @Lucifer3224 we should close this now, right?

yup yup, I'm already working on it for a new pr

PR #318, opened 2026-02-17
refactor(canvas): decompose FlowCanvas into focused hooks and components

ashmod:
Extracts hooks and components from the monolithic FlowCanvas, and applies some cleanup for maintainability.


unauthorised-401 (commented):
## Nice extraction overall
##### One small catch
In useGraphClipboard.ts:87, exportToClipboard calls navigator.clipboard.writeText without a permission/try-catch guard (unlike copySelection above).

Could we mirror the same error handling here? This would ensure clipboard failures show a user-friendly toast instead of throwing silently.

PR #351, opened 2026-02-21
refactor(frontend): shared confirmation dialog

ZeynatGhallab:
Implementing shared confirmation dialog component.


ashmod:
Thanks! Let's fix the failing typechecks and this should be good to merge.

PR #361, opened 2026-02-23
refactor(api): seperate setup auth routes

Lucifer3224:
### Updated files related to the first sign up

-  Moved the first-time setup routes into a new dedicated **setup** module with its own router, service, and schemas
-  Renamed the endpoints to better reflect their purpose: **/setup/status** and **/setup/initialize**
-  Registered the new **setup_router** in **app.py** alongside the existing **auth_router**
-  Updated **openapi.yaml** to reflect the new endpoint
-  Updated the frontend autogenerated client files
-  Updated the frontend API wrapper **auth.ts** to call the new SDK functions 
-  Migrated all 24 setup-related tests from **test_first_admin_signup.py** into a new **test_setup.py** file with paths updated to the new endpoints, then deleted the old test file

### DIFF
--- a/services/api/test/setup/test_setup.py
+++ b/services/api/test/setup/test_setup.py
@@ -15,10 +15,8 @@ async def test_first_time_setup_status_when_no_users(client: AsyncClient):
     assert response.status_code == 200
     data = response.json()
     assert data["success"] is True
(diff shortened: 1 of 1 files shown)


Lucifer3224:
check

PR #370, opened 2026-02-26
fix(wf-role): add wf role to the GET /workflows/

Spafic:
Reviewed as requested in #258 by @ashmod. The backend now includes the user’s role in the response.


ashmod:
i think this was rebased off Shahd's pr, could you rebase from main or open a new one off it? so we can merge this before the frontend's.

PR #371, opened 2026-02-26
feat(docs): integrate Nextra documentation with app router

Lucifer3224:
### Summary of what I've done:

        - Integrated with Nextra v4
        - Created new App Router documentation structure at src/app/docs/
        - Added 2 new quick access cards in the docs home page for Scryb and Smith
        - Overrode Nextra's default layout to integrate the footer and refined navigation.
        - Added theme switcher to the doc's side bar
        - Added direct access links for providing feedback and to the content pages.

### DIFF
--- a/apps/web/src/components/layout/DocsNavbar.tsx
+++ b/apps/web/src/components/layout/DocsNavbar.tsx
@@ -7,7 +7,7 @@ export function DocsNavbar() {
   return (
     <nav className="flex items-center justify-between w-full h-16 px-6 bg-background/80 backdrop-blur-md border-b border-border/40">
       <div className="flex items-center gap-6">
-        <Logo variant="glyph" href="/create/docs" className="h-6 w-6" />
+        <Logo variant="glyph" href="/" className="h-6 w-6" />
       </div>
 
       <div className="flex-1" />


Lucifer3224:
That worked! Thanks a lot for the suggestion🫡

unauthorised-401:
I looked into the build error and found out we can fix it just by adding a few lines to our next.config.ts!

This means we don't need to revert to Webpack, and we get to keep Turbopack's blazing-fast local startup times. Because Turbopack is stricter with module resolution, it just needs a direct alias to find Nextra's virtual MDX file @Lucifer3224 @ashmod 

```typescript
// next.config.ts
const nextConfig: NextConfig = {
  // ... our existing config
  turbopack: {
    resolveAlias: {
      'next-mdx-import-source-file': './mdx-components.tsx',
    },
  },
};
```

ashmod:
Nice work! I really like some of the parts especially the copy MD and the LLM buttons.

I still haven't gotten around to reviewing the actual code, but a few points to keep in mind based on the app as I experimented with it:

- The documentation content (including TOC) should ideally be removed (we can add a placeholder for now). This is because we want the documentation to tick off a few certain criteria, so we'd likely have to hold meetings over the actual documentation content - so let's keep this PR scoped only to the Nextra integration.

- There are a few bugs with the UI (in addition to the failing lint action). Please thoroughly re-test it. These are a couple of the issues I faced on running it:

<details>
<summary><strong>Console Error Log</strong></summary>

```md
## Error Type
Console Error

## Error Message
In HTML, <a> cannot be a descendant of <a>.
This will cause a hydration error.

  ...
    <RedirectBoundary>
      <RedirectErrorBoundary router={{...}}>
        <InnerLayoutRouter url="/docs" tree={[...]} params={{}} cacheNode={{rsc:<Fragment>, ...}} segmentPath={[...]} ...>
          <SegmentViewNode type="page" pagePath="docs/page....">
            <SegmentTrieNode>
            <MDXWrapper>
              <wrapper>
                <div className="x:mx-auto ...">
                  <TOCProvider value={[...]}>
                    <Sidebar>
                    <ClientWrapper metadata={{title:"Run...", ...}} sourceCode={"# Rune D..."}>
                      <nav className="nextra-toc..." aria-label="table of c...">
                        <TOC filePath="src/app/do..." pageTitle="Rune Docum...">
                          <div className="x:grid x:g...">
                            <p>
                            <ul>
                            <div className="x:grid x:g...">
                              <Anchor className="x:text-xs ..." href={"https://..."}>
>                               <a
>                                 href={"[https://github.com/rune-org/rune/issues/new?title=Feedback%20for%20%E2%80%9C](https://github.com/rune-org/rune/issues/new?title=Feedback%20for%20%E2%80%9C)..."}
>                                 target="_blank"
>                                 rel="noreferrer"
>                                 className="x:focus-visible:nextra-focus x:text-xs x:font-medium x:transition x:text-..."
>                               >
>                                 <a href="/not-found">
                              ...
                      ...
          ...
        ...



    at a (unknown:0:0)
    at DocsLayout (src/app/docs/layout.tsx:21:18)

## Code Frame
  19 |       editLink={<a href="/not-found">Edit this page</a>}
  20 |       feedback={{
> 21 |         content: <a href="/not-found">Question? Give us feedback →</a>,
     |                  ^
  22 |         labels: 'feedback',
  23 |       }}
  24 |     >

Next.js version: 16.1.6 (Turbopack)
```

</details>

<details>
<summary><strong>Recoverable Error Log</strong></summary>

```md
## Error Type
Recoverable Error

## Error Message
Hydration failed because the server rendered HTML didn't match the client. As a result this tree will be regenerated on the client. This can happen if a SSR-ed Client Component used:

- A server/client branch `if (typeof window !== 'undefined')`.
- Variable input such as `Date.now()` or `Math.random()` which changes each time it's called.
- Date formatting in a user's locale which doesn't match the server.
- External changing data without sending a snapshot of it along with the HTML.
- Invalid HTML tag nesting.

It can also happen if the client has a browser extension installed which messes with the HTML before React loaded.

[https://react.dev/link/hydration-mismatch](https://react.dev/link/hydration-mismatch)

  ...
    <MobileNav>
      <aside className={"nextra-m..."}>
        ...
          <Ae>
            <c value={2}>
              <div className="nextra-sea...">
                <Ro spellCheck={false} autoComplete="off" type="search" ref={{current:null}} className={function _temp4} ...>
                  <input
                    spellCheck={false}
                    autoComplete="off"
                    className={"x:rounded-lg x:px-3 x:py-2 x:transition-all x:w-full x:md:w-64 x:text-base x:leading-..."}
                    value=""
                    placeholder="Search documentation…"
+                   id="headlessui-combobox-input-_R_mqrkndlb_"
-                   id="headlessui-combobox-input-_R_2rbekndlb_"
                    role="combobox"
                    type="search"
                    aria-controls={undefined}
                    aria-expanded={false}
                    aria-activedescendant={undefined}
                    aria-labelledby={undefined}
                    aria-describedby={undefined}
                    aria-autocomplete="list"
                    defaultValue={undefined}
                    disabled={undefined}
                    autoFocus={false}
                    onFocus={function onFocus}
                    onBlur={function onBlur}
                    onCompositionStart={function onCompositionStart}
                    onCompositionEnd={function onCompositionEnd}
                    onKeyDown={function onKeyDown}
                    onChange={function onChange}
                    onPointerEnter={function onPointerEnter}
                    onPointerLeave={function onPointerLeave}
                    ref={function o.useCallback}
                    data-headlessui-state=""
                  >
                ...
              ...
        <Menu>
        ...
          <gt value="light" onChange={function V.useCallback[f]}>
            <Ae>
              <c value={2}>
                <Et title="Change theme" className={function t3}>
                  <button
                    title="Change theme"
                    className="x:cursor-pointer x:h-7 x:rounded-md x:px-2 x:text-xs x:font-medium x:transition-colors ..."
+                   id="headlessui-listbox-button-_R_rerkndlb_"
-                   id="headlessui-listbox-button-_R_3drekndlb_"
                    type="button"
                    aria-haspopup="listbox"
                    aria-controls={undefined}
                    aria-expanded={false}
                    aria-labelledby={undefined}
                    aria-describedby={undefined}
                    disabled={undefined}
                    autoFocus={false}
                    onKeyDown={function onKeyDown}
                    onKeyUp={function onKeyUp}
                    onKeyPress={function onKeyPress}
                    onPointerDown={function onPointerDown}
                    onClick={function onClick}
                    onFocus={function onFocus}
                    onBlur={function onBlur}
                    onPointerEnter={function onPointerEnter}
                    onPointerLeave={function onPointerLeave}
                    onPointerUp={function onPointerUp}
                    ref={function o.useCallback}
                    data-headlessui-state=""
                  >
                ...
    ...
      <InnerScrollAndFocusHandler segmentPath={[...]} focusAndScrollRef={{apply:false, ...}}>
        <ErrorBoundary errorComponent={undefined} errorStyles={undefined} errorScripts={undefined}>
          <LoadingBoundary name="docs/" loading={null}>
            <HTTPAccessFallbackBoundary notFound={undefined} forbidden={undefined} unauthorized={undefined}>
              <RedirectBoundary>
                <RedirectErrorBoundary router={{...}}>
                  <InnerLayoutRouter url="/docs" tree={[...]} params={{}} cacheNode={{rsc:<Fragment>, ...}} ...>
                    <SegmentViewNode type="page" pagePath="docs/page....">
                      <SegmentTrieNode>
                      <MDXWrapper>
                        <wrapper>
                          <div className="x:mx-auto ...">
                            <TOCProvider value={[...]}>
                              <Sidebar>
                                <aside
+                                 id="_R_pbnbkndlb_"
-                                 id="_R_35etekndlb_"
                                  className="nextra-sidebar x:print:hidden x:transition-all x:ease-in-out x:max-md:hid..."
                                >
                                  <div>
                                  <div className="x:sticky x...">
                                    <LocaleSwitch>
                                    ...
                                      <Et title="Change theme" className={function t3}>
                                        <button
                                          title="Change theme"
                                          className="x:cursor-pointer x:h-7 x:rounded-md x:px-2 x:text-xs x:font-mediu..."
+                                         id="headlessui-listbox-button-_R_6tapbnbkndlb_"
-                                         id="headlessui-listbox-button-_R_rlb5etekndlb_"
                                          type="button"
                                          aria-haspopup="listbox"
                                          aria-controls={undefined}
                                          aria-expanded={false}
                                          aria-labelledby={undefined}
                                          aria-describedby={undefined}
                                          disabled={undefined}
                                          autoFocus={false}
                                          onKeyDown={function onKeyDown}
                                          onKeyUp={function onKeyUp}
                                          onKeyPress={function onKeyPress}
                                          onPointerDown={function onPointerDown}
                                          onClick={function onClick}
                                          onFocus={function onFocus}
                                          onBlur={function onBlur}
                                          onPointerEnter={function onPointerEnter}
                                          onPointerLeave={function onPointerLeave}
                                          onPointerUp={function onPointerUp}
                                          ref={function o.useCallback}
                                          data-headlessui-state=""
                                        >
                                      ...
                                    <Button aria-expanded={true} aria-controls="_R_pbnbkndlb_" title="Collapse s..." ...>
                                      <v className={function t2} aria-expanded={true} aria-controls="_R_pbnbkndlb_" ...>
                                        <button
                                          className="x:transition x:cursor-pointer x:rounded-md x:p-2 x:text-gray-600 ..."
                                          aria-expanded={true}
+                                         aria-controls="_R_pbnbkndlb_"
-                                         aria-controls="_R_35etekndlb_"
                                          title="Collapse sidebar"
                                          type="button"
                                          disabled={undefined}
                                          autoFocus={false}
                                          onClick={function onClick}
                                          onFocus={function onFocus}
                                          onBlur={function onBlur}
                                          onPointerEnter={function onPointerEnter}
                                          onPointerLeave={function onPointerLeave}
                                          onPointerDown={function onPointerDown}
                                          onPointerUp={function onPointerUp}
                                          ref={null}
                                          data-headlessui-state=""
                                        >
                              ...
                    ...
                  ...
        ...



    at a (unknown:0:0)
    at DocsLayout (src/app/docs/layout.tsx:21:18)

## Code Frame
  19 |       editLink={<a href="/not-found">Edit this page</a>}
  20 |       feedback={{
> 21 |         content: <a href="/not-found">Question? Give us feedback →</a>,
     |                  ^
  22 |         labels: 'feedback',
  23 |       }}
  24 |     >

Next.js version: 16.1.6 (Turbopack)
```

</details>

- Finally, light theme seems to be broken on my end:
image

ashmod:
I recently added a formatting check on the frontend, please run 

```bash
make web-format
```
locally to fix any lint issues

ashmod:
@Lucifer3224 Light mode is ready now, so feel free to continue working on this one.

Lucifer3224:
I've merged the light theme updates, so it worked automatically.
image

Also, the errors you were facing finally appeared on my end, so I fixed it and now the feedback routes to rune issues with an auto-generated title: `feedback for "{name of the section where it was opened when feedback was clicked}"`.
image

For `"Edit this page"`, it will work once we start adding the MDX files.

I guess everything is now working perfectly and ready for the content.

ashmod:
This looks really nice and clean, thank you!

I have a few UI-related concerns, highlighted below:

runedocsfb

Suggestion: the top part's breathing area can include a search bar, it can be no-op for now while we decide on how to get it as efficient and fast as possible. It can also include a couple other stuff like language, and other quick shortcuts. Example: https://www.mintlify.com/docs , and a more minimal example: https://opencode.ai/docs

Let me know if you need any further clarification on any of those points!

apps/web/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev --turbopack",
-    "build": "next build --turbopack",
+    "dev": "next dev",
+    "build": "next build",
```
ashmod, line 7: I completely forgot what the situation was here, but did we decide to disable turbopack? I remember @unauthorised-401 proposed a solution that kept turbopack usable with nextra. 

I honestly don't remember so refresh my memory on this one :D

apps/web/src/app/docs/layout.tsx
```diff
@@ -0,0 +1,24 @@
+import { Layout } from "nextra-theme-docs";
+import { getPageMap } from "nextra/page-map";
+import "nextra-theme-docs/style.css";
+
+export const dynamic = "force-dynamic";
```
ashmod: This makes every `/docs/**` route render on demand instead of being statically generated, which hurts cacheability and adds avoidable latency for checked-in MDX docs. We can safely remove this.

apps/web/src/app/globals.css
```diff
@@ -112,6 +112,11 @@
     color: var(--foreground);
   }
 
+  /* Hide the # anchor links on Nextra doc headings (keep IDs for TOC sidebar) */
+  .subheading-anchor {
+    display: none !important;
+  }
```
ashmod: this class name is too generic, consider scoping it under `.nextra-content` for example to avoid collisions.

apps/web/src/app/globals.css
```diff
@@ -121,6 +126,12 @@
     background-color: hsl(222 47% 11% / 0.62);
     color: hsl(0 0% 100%);
   }
+
+  /* Hide the external link icons in Nextra */
+  nav.nextra-toc a svg,
+  .nextra-sidebar-footer a svg {
+    display: none !important;
+  }
```
ashmod: If we ever decide to use svg icons inside the TOC, this might block it if I'm not mistaken. I think one of the reasons we could use icons is to group sections and have them organized more attractively.

Lucifer3224:
> This looks really nice and clean, thank you!
> 
> I have a few UI-related concerns, highlighted below:
> 
> runedocsfb
> Suggestion: the top part's breathing area can include a search bar, it can be no-op for now while we decide on how to get it as efficient and fast as possible. It can also include a couple other stuff like language, and other quick shortcuts. Example: https://www.mintlify.com/docs , and a more minimal example: https://opencode.ai/docs
> 
> Let me know if you need any further clarification on any of those points!


You're welcome, glad you like it💜
Regarding the breathing room at the top, I totally agree, but I kept it minimal for now just to make sure the base spacing feels right (from my POV), especially since we can easily add the search bar and shortcuts into a header without messing with the layout.

Also, for the sidebar, it feels "outside" Rune right now because that's Nextra’s default "whole world" behavior for standalone docs. It doesn’t automatically connect to the rest, but adding that header, as I mentioned, will solve this too by bringing in the Rune logo and a path back home so it feels fully integrated.

And about the theme switcher, Nextra's default behavior is to have it follow the navigation links, which is why it looks a bit off sometimes on pages with only a few sections, but I can override it to dock it permanently at the bottom.

Could you let me know if you prefer implementing a header, separate from the Home one, (I honestly recommend this) or should I just proceed with the refinements you mentioned?

apps/web/src/app/globals.css
Lucifer3224: I tested it, and you are right. The SVGs became hidden, so I fixed and tested it again, and it's now working

apps/web/src/app/globals.css
Lucifer3224: Done

apps/web/src/app/docs/layout.tsx
Lucifer3224: Done

apps/web/package.json
Lucifer3224: Turbopack is working. Back then, we were forcing Webpack with the `--webpack` flag because Nextra was having issues. We then fixed that by removing the flag and adding the `next-mdx-import-source-file alias` to the turbopack block in `next.config.ts` , the solution which @unauthorised-401 proposed.

ashmod: Nice, so the --turbopack omission here is intentional and not out of date, correct?

ashmod:
A separate header sounds really nice, let's go in that direction! Let's have it minimal and allow users to go back and forth between the main app and the docs pages.

apps/web/package.json
Lucifer3224: Yup

ashmod:
Looks nice, the "back to rune" text is kind of redundant since clicking the logo should just redirect users back. Also let's make the logo a bit smaller/more compact.

As for the theme toggler, it's common to have a toggle for the docs page (separate from the footer's) so we'll keep it like that. The placement is justified to have the toggle available everywhere in the app regardless of the page content (for example, homepage has a navbar but the main app does not, and the docs page has a separate one.)

But theming defaults to system anyway and users are able to see it clearly on the homepage when they visit the platform, from their profile settings when they visit the app, and now in the docs sidebar too. I think that is sufficient.

apps/web/src/components/layout/DocsNavbar.tsx
```diff
@@ -7,15 +7,7 @@ export function DocsNavbar() {
...
-        <Logo variant="glyph" href="/create/docs" className="h-8 w-8" />
-        <div className="h-4 w-px bg-border/60" />
-        <Link
-          href="/create/docs"
-          className="text-sm font-medium text-muted-foreground hover:text-foreground transition-colors flex items-center gap-2"
-        >
-          <span className="hidden sm:inline">Back to Rune</span>
-          <span className="sm:hidden">Exit</span>
-        </Link>
+        <Logo variant="glyph" href="/create/docs" className="h-6 w-6" />
```
ashmod: this should point back to "/" (the rune homepage)

If you'd like to have users to navigate back to the Docs home, you could add some shortcuts in the navbar like "Home" , "Changelog" ,... etc. (i.e. common hot spots.) even though currently the TOC should be enough.

But make sure that the Docs navigation is back to `/docs` and **not** `/create/docs` because `/create` is guarded by authentication - we wouldn't want an unauth'd user to have to authenticate to access docs.

apps/web/src/app/docs/layout.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { DocsNavbar } from "@/components/layout/DocsNavbar";
+import { Footer } from "@/components/layout/Footer";
+
+export default async function DocsLayout({ children }: { children: React.ReactNode }) {
+  const pageMap = await getPageMap("/docs");
+
+  return (
+    <Layout
+      pageMap={pageMap}
+      docsRepositoryBase="https://github.com/rune-org/rune"
```
ashmod: Just noticed this, but `docsRepositoryBase` should include the docs subdirectory in our repo. Right now Nextra will build "Edit this page" links relative to `https://github.com/rune-org/rune`, so a page like `/docs/getting-started` ends up pointing at a non-existent root-level path instead of `apps/web/src/app/docs/getting-started/page.mdx`. Setting the base to `https://github.com/rune-org/rune/tree/main/apps/web/src/app/docs` should make those links resolve to the actual source files.

apps/web/src/components/layout/DocsNavbar.tsx
Lucifer3224: but this isn't possible to happen becaue the rest of the pages, as you said, are guarded by auth, so the user will just be redirected to the sign-in form.
You are authenticated? Move back to where you were before docs
You are not? Kick

apps/web/src/app/docs/layout.tsx
Lucifer3224: you're right, I've fixed that.

apps/web/src/components/layout/DocsNavbar.tsx
ashmod: Wdym 😂
Docs are not guarded by auth and should not be, they live in /docs, and they can actually access it from the homepage's "Docs" in the navbar. the /create/docs is just a convenient shortcut for those inside the rune app already.

The homepage requires no auth (the landing page with our marketing elements.) so this should be the page users are redirected to when clicking on the Rune icon.

For example, https://geminicli.com/docs/cli/tutorials/memory-management/ clicking the gemini cli icon on the topleft of this page navigates you back to the geminicli marketing homepage, not the docs.

Lmk if this makes sense now or if something is not clicking for you yet.

Lucifer3224: Of course, I didn’t mean that Home or Docs are guarded 😂😭
image

I understand your point, you’re saying the Docs icon should point back to the Rune homepage not to "/create/docs" since that route is part of the authenticated app, and I agree with that.

What I was referring to is the navigation behavior once the user is already inside Docs, which is publicly accessible, and this is the current behavior, and clicks the icon while in there, there are two scenarios:

1. If the user is authenticated, they should be taken back to their previous context inside the app (e.g. "/create/docs").
2. If the user is not authenticated, they will naturally be redirected to the sign-in page when trying to access the app routes.

The difference compared to something like Gemini CLI is that their docs are fully public and stateless, so the logo always just routes back to the homepage ("/") with no concept of an authenticated app context or previous state. In our case, we have both a public docs space and an authenticated app, which introduces this contextual navigation behavior.

ashmod: Ok so I don't think my point is fully coming through here. But imagine this, a user just downloaded rune, went into the homepage, finds "Docs" and clicks it, so they go and read on the app. They then decide to go back to the homepage to see what else was in store, but when they click the Rune icon they find a sign in screen even though they never requested to sign in in the first place, and most importantly, signing in in that specific case won't work at all because the user should sign up (separate problem for another time, but still shouldn't be the best behavior here).

So, to simplify behavior, we should have docs, as you called it, more "stateless".. we want it to feel like an isolated experience from the rune app, but at the same time have a convenient shortcut from the app shell for logged in users. So by default, docs should just be a standalone experience and not feel tightly coupled to the app's actual content. Does that make more sense?

PR #400, opened 2026-03-07
fix(canvas): handle RTES connection loss during workflow execution

ashmod:
This change improves the canvas execution UX when the realtime execution service is unavailable. Instead of leaving workflows looking indefinitely healthy and stuck in a generic running state, the UI now distinguishes between execution progress and live connection health.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,5 @@ nginx/ssl/
 
 # Cursor
 .cursor/
+
+.claude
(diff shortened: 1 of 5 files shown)


ashmod:
Dismiss has been renamed to Halt. As for the UX problem, it's not of high priority at the moment, but it should be easier to handle once we have the Executions API middleware @unauthorised-401 working on is ready, we can give it another look then.

PR #403, opened 2026-03-07
feat(web): add quick actions and export to workflows table

### LINKED ISSUE #389: Refactor /workflows user actions
Currently our /create/workflows page has the actions displayed poorly UX-wise, for example, to delete or run a workflow a user would have to click on the "..." then select the option, that's one too many clicks. We should look into the possibility of refactoring the actions to include "quick actions" for stuff like this.

We could even refactor the whole table component's ui if quick actions are not suitable ui-wise with the current component.

Additionally, users currently have no way to export a workflow (to json) from the workflows list page, we should add that action to the list of available actions.

DopeBiscuit:
## Summary

Adds **quick actions** (Run, Export, Delete) to the workflows table and **export workflow to JSON** from the list, so common actions are one click and workflows can be exported without opening the canvas.

Fixes #389.

## Changes

- **Quick actions:** Inline icon buttons on each workflow row — **Run** (Play), **Export** (Download), **Delete** (Trash) — with tooltips. One click for each action instead of opening the "..." menu.
- **Export to JSON:** New action that fetches the workflow via `getWorkflowById`, strips credentials with `stripCredentials`, and downloads `workflow-{name}-{id}.json` with `name`, `description`, and `workflow_data`. Available as quick action and in the "..." menu.
- **"..." menu:** Unchanged set of actions (Open in Canvas, Rename, Activate/Deactivate, Run, Share, Delete) plus **Export to JSON**. Run, Export, and Delete stay in the menu as well.
- **Permissions:** Quick actions respect existing checks: `canExecuteWorkflow` (Run), `canViewWorkflow` (Export), `canDeleteWorkflow` (Delete).

## Effects

- Fewer clicks for Run, Export, and Delete from the workflows list.
- Workflows can be exported to JSON from the list without opening the canvas.
- Export output is safe to share (credentials stripped from node data).

## Test checklist

- [x] At `/create/workflows`, each row shows Run, Export, Delete icons and "..." with tooltips.
- [x] **Run:** Click Play → workflow is queued; success toast.
- [x] **Export:** Click Download → JSON file downloads; contains `name`, `description`, `workflow_data`; node credentials stripped.
- [x] **Delete:** Click Trash → confirmation dialog; cancel or confirm works.
- [x] Export from "..." menu works and downloads same JSON.
- [x] Run/Export/Delete hidden or disabled when user lacks permission (e.g. viewer).

## Notes

- Toast wording for Run ("Workflow successfully executed.") and duplicate/misleading execution toasts (e.g. when RTES WebSocket fails) are tracked in a separate improvement issue; not changed in this PR.
- When running a workflow from the workflows menu, the missing execution service toasts don't show up, this needs to be checked.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -17,7 +17,7 @@ http {
 
     server {
         listen 80;
-        server_name localhost;
+        server_name _;
 
         # Redirect all HTTP requests to HTTPS
         return 301 https://$host$request_uri;
@@ -25,7 +25,7 @@ http {
 
     server {
         listen 443 ssl;
-        server_name localhost;
+        server_name _;
 
         ssl_certificate /etc/nginx/ssl/server.crt;
         ssl_certificate_key /etc/nginx/ssl/server.key;


apps/web/src/components/workflows/WorkflowsTable.tsx
```diff
@@ -191,6 +211,50 @@ export function WorkflowsTable() {
...
+        return;
+      }
+      const detail = response.data.data;
+      const rawData = detail.workflow_data as Record<string, unknown> | undefined;
+      const hasGraph =
+        rawData &&
+        Array.isArray(rawData.nodes) &&
+        Array.isArray(rawData.edges);
+      const workflowData = hasGraph
+        ? stripCredentials(rawData as RFGraph)
```
ashmod: `detail.workflow_data` is the persisted workflow DSL, but `stripCredentials` only sanitizes the canvas graph shape by removing `node.data.credential`. The DSL still stores secrets in `node.credentials`, so exporting this object here can still write live credentials into the downloaded JSON for configured nodes.

ashmod (approved):

Issue #391, opened 2026-03-04
Allow multi-selection for workflow actions


ashmod:
We should allow users to multi-select workflows and apply relevant actions (e.g. batch delete, run, export, etc.).

Exporting is still blocked by #389 but when it's ready a batch export should ideally export a zip including all workflows as JSONs.

ashmod:
@Spafic you are good to go on this one, #389 was merged.

PR #416, opened 2026-03-10
feat(api): add bulk workflow operations endpoint

Spafic:
This PR implements:
- Bulk workflow operation endpoint
- Centralized workflow run logic

Required for #391


services/api/src/workflow/router.py
```diff
@@ -292,7 +429,6 @@ async def delete_workflow(
 @router.post("/{workflow_id}/run", response_model=ApiResponse[str])
 @require_workflow_permission("execute")
 async def run_workflow(
-    db: DatabaseDep,
```
ashmod: same here

Spafic:
> Looks like a merge commit discarded recent changes on main. Please doublecheck.

Reverify the changes with the services file as I have refactored the /run logic to be centralized there using`queue_workflow_execution` function

PR #429, opened 2026-03-12
feat(frontend): implement SAML callback handling and exchange code functionality

Spafic:
This PR implements the exchange handling between the IdP and SP. After successful login at the IdP, the user is redirected to RUNE with an exchange key, eliminating the need to expose tokens in the URL. The exchange key is then sent to the `/exchange` endpoint to obtain the access and refresh tokens. Additionally, a loading state is displayed until the credentials are resolved and the SAML user is authenticated, after which the user is redirected to the home page.


ashmod:
Please add a pr description to this.

PR #512, opened 2026-04-13
feat(web): add once-in-a-lifetime admin visibility warning for users

### LINKED ISSUE #508: User not informed that admin can see their data
Users are not informed that the admin can view their credentials, workflows, and other data. This should be probably hinted at somewhere.

Lucifer3224:
I've finished implementing the admin visibility warning modal.

I took a frontend approach using `[localStorage]`. It's a "once in a lifetime" that appears the first time a user signs in after they change their temporary password and included a couple of safeguards:

The logic checks the user's role, so it strictly targets regular users and bypasses admins.
The storage key is tied to the specific user ID. This ensures that if multiple people sign in from the same pc and browser, they will each properly receive the notice on their respective accounts.
image

resolves #508

### DIFF
--- a/apps/web/src/components/auth/AdminVisibilityModal.tsx
+++ b/apps/web/src/components/auth/AdminVisibilityModal.tsx
@@ -1,0 +1,52 @@
+"use client";
+
+import { useEffect, useState } from "react";
+import { Dialog, DialogContent, DialogDescription, DialogFooter, DialogHeader, DialogTitle } from "@/components/ui/dialog";
+import { Button } from "@/components/ui/button";
+import { useAuth } from "@/lib/auth";
+
+const STORAGE_KEY_PREFIX = "admin_visibility_confirmed_";
+
+export function AdminVisibilityModal() {
+  const [open, setOpen] = useState(false);
+  const { state } = useAuth();
+  
+  // Create a user-specific storage key
+  const storageKey = state.user?.id ? `${STORAGE_KEY_PREFIX}${state.user.id}` : null;
+
+  useEffect(() => {
+    // Determine if user is not loaded or is an admin
+    if (!state.user || state.user.role === "admin" || !storageKey) {
+      return;
+    }
+
+    // Only access localStorage on the client to avoid hydration mismatch
+    const hasConfirmed = localStorage.getItem(storageKey);
+    if (!hasConfirmed) {
+      setOpen(true);
+    }
+  }, [state.user, storageKey]);
+
+  const handleConfirm = () => {
+    if (storageKey) {
+      localStorage.setItem(storageKey, "true");
+    }
+    setOpen(false);
+  };
+
+  return (
+    <Dialog open={open} onOpenChange={setOpen}>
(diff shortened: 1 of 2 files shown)


ashmod:
Nice work! We need to make sure this only displays to "Viewer" roles which I can see is done so this looks great :D

Will review this shortly.

apps/web/src/components/auth/AdminVisibilityModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+export function AdminVisibilityModal() {
+  const [open, setOpen] = useState(false);
+  const { state } = useAuth();
+
+  // Create a user-specific storage key
+  const storageKey = state.user?.id ? `${STORAGE_KEY_PREFIX}${state.user.id}` : null;
+
+  useEffect(() => {
+    // Determine if user is not loaded or is an admin
+    if (!state.user || state.user.role === "admin" || !storageKey) {
```
ashmod, line 28:
```suggestion
    if (!state.user || state.user.role === "admin" || state.user.must_change_password || !storageKey) {
```
We shouldn't show this while the user is still in the forced password-change flow because the form would be occupying the screen.

apps/web/src/components/auth/AdminVisibilityModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const handleConfirm = () => {
+    if (storageKey) {
+      localStorage.setItem(storageKey, "true");
+    }
+    setOpen(false);
+  };
+
+  return (
+    <Dialog open={open} onOpenChange={setOpen}>
```
ashmod: This keeps the dialog dismissible through the normal Radix close paths. Clicking the overlay, pressing Escape, or using the built-in close button will close it, but nothing is persisted to localStorage. So the user would get the alert again on the next visit.

ashmod (commented):
Nice work, I think we should add a todo somewhere to have this as a per-user flag instead of persisting to localStorage, since visiting in another browser, incognito, or a different environment would all re-trigger it.

apps/web/src/components/auth/AdminVisibilityModal.tsx
Lucifer3224: But I don't think it's possible to show up during the password change because the modal lives inside `AppShell` and the app redirects users to the change-password page (which doesn't use `AppShell` at all) before `AppShell` even mounts.

apps/web/src/components/auth/AdminVisibilityModal.tsx
Lucifer3224: Good catch, I updated it.

Lucifer3224:
Great minds! I actually thought about adding it to the DB originally too but I decided to stick with localStorage for noe because it's a minor UI pop-up, but you're completely right about the trade-off so, I just pushed a TODO comment so we can circle back and add it to the user model later!

ashmod (approved):
LGTM!

PR #520, opened 2026-04-15
fix(initials): implement getInitials utility function

### LINKED ISSUE #510: Avatar UI shows too many initials
The avatar circle currently renders the first letter of every word in the user's display name. For names with multiple parts (e.g. "John Doe Something Something"), this produces a cluttered "JDSS" that overflows the small circle.

We should have it display at most 2 initials: the first letter of the first name and the first letter of the last name.

Spafic:
- closes #510 
- Fixed and centralized initials for future usage

### DIFF
--- a/apps/web/src/lib/initials.ts
+++ b/apps/web/src/lib/initials.ts
@@ -8,7 +8,10 @@
  */
 export const getInitials = (name: string): string => {
   if (!name) return "U";
-  const parts = name.split(" ").filter((part) => part.length > 0);
+  const parts = name
+    .trim()
+    .split(/\s+/)
+    .filter((part) => part.length > 0);
   if (parts.length === 0) return "U";
   if (parts.length === 1) return parts[0].charAt(0).toUpperCase();
   return (parts[0].charAt(0) + parts[parts.length - 1].charAt(0)).toUpperCase();
--- a/apps/web/src/lib/utils/initials.ts
+++ b/apps/web/src/lib/utils/initials.ts
@@ -1,0 +1,15 @@
+/**
+ * Returns the initials of a given name.
+ * - If the name is empty, returns "U".
+ * - For names with multiple words, returns the first letter of the first and last name.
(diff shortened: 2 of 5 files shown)


ashmod:
Nice, can you add tests?

apps/web/src/lib/initials.ts
ashmod, line 1: we can move that outside of utils for now since we're following that convention where utility methods and their tests live inside `lib/`.

apps/web/src/lib/utils/initials.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Returns the initials of a given name.
+ * - If the name is empty, returns "U".
+ * - For names with multiple words, returns the first letter of the first and last name.
+ * - For single-word names, returns just the first letter.
+ * @param name The full name string.
+ * @returns The initials (up to 2 uppercase letters), or "U" if not available.
+ */
+export const getInitials = (name: string): string => {
+  if (!name) return "U";
+  const parts = name.split(" ").filter((part) => part.length > 0);
```
ashmod: `split(" ")` would miss tabs and multi-space edge cases, consider replacing with
```ts
name.trim().split(/\s+/)
```
which also removes the need for the `.filter`

Spafic: Nice catch, Thank u ❤️

Spafic: The filter part is important; if the name is "         ", the avatar will be blank
image

Surprisingly, backend accepts empty strings 🥳

ashmod: Well.. that shouldn't be allowed :D

You could open an issue on that backend problem.

Spafic: I will and keep the filter as a final frontend validation :)

ashmod (approved):
LGTM

PR #523, opened 2026-04-15
fix(profile): refactor profile update logic to refetch user data instead of logging out

### LINKED ISSUE #492: Changing username logs the user out
Updating the user's display name via the profile page causes an immediate logout. The name change does persist (visible after re-login), but the forced logout is disruptive.

#### Steps to Reproduce

1. Go to profile/settings.
2. Change the display name.
3. Save.

#### Expected Behavior

Name updates without logging the user out. The new name is reflected immediately.

#### Actual Behavior

The user is logged out. After re-login, the new name is visible.

Spafic:
- closes #492 
- Replaced logout calling with a refeatch profile

### DIFF
--- a/apps/web/src/app/(app)/profile/page.tsx
+++ b/apps/web/src/app/(app)/profile/page.tsx
@@ -47,7 +47,7 @@ export default function ProfilePage() {
   const [error, setError] = useState<string | null>(null);
   const [isPasswordDialogOpen, setIsPasswordDialogOpen] = useState(false);
   const [dialogResetKey, setDialogResetKey] = useState(0);
-  const { state, logout, refetchProfile } = useAuth();
+  const { state, logout, refetchProfile, refresh } = useAuth();
 
   const user = state.user as UserResponse | null;
   const roleLabel = user?.role ? user.role.charAt(0).toUpperCase() + user.role.slice(1) : "User";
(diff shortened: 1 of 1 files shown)


apps/web/src/app/(app)/profile/page.tsx
```diff
@@ -74,9 +74,11 @@ export default function ProfilePage() {
       await updateMyProfile({
         name: result.data.name,
       });
-
-      // Log out user after successful update
-      await logout();
+      // Refetch profile to sync state across the app
+      await refetchProfile();
```
ashmod, line 80: `refetchProfile()` calls `GET /profile/me`, which resolves user data from [`decode_access_token()`](https://github.com/rune-org/rune/blob/55442446b5ff3a6ed9b537895e0b8ba3d5c895b4/services/api/src/core/token.py#L56-L85) - the user is reconstructed from the JWT payload, not from the database. Since `PUT /profile/me` doesn't issue a new token, the access token still contains the old name which means the name actually doesn't update unless the token refreshes.

So, if you try to update the name in that page, it would update in the backend properly **but the frontend would not reflect that until the token is refreshed.**

Possible fixes:
1. Have `PUT /profile/me` return the updated user and set a new access token cookie in the response, then use that response client-side instead of refetching.
2. Trigger a token refresh (`await refresh()`) after the profile update so a new JWT is minted with the updated name.
3. Have `GET /profile/me` read from the database instead of the token (changes the endpoint semantics though).

apps/web/src/app/(app)/profile/page.tsx
```diff
@@ -74,9 +74,11 @@ export default function ProfilePage() {
...
+      // Close edit mode
+      setEditingField(null);
+      setIsSaving(false);
     } catch (apiError) {
       setError(getErrorMessage(apiError));
       setIsSaving(false);
```
ashmod, line 86: this is now duplicated in both `try` and `catch` blocks. Consider cleaning up with a `finally`.

ashmod (approved):
LGTM!

PR #527, opened 2026-04-16
feat: Implement template deletion for owners and admins

### LINKED ISSUE #491: No option to delete templates
There is no delete action available for templates. Once created, templates cannot be removed.

Lucifer3224:
The template deletion feature is now implemented and tested. I've updated the backend schema and service to allow the admin also to delete a template even if he is not the creator, and added the delete button, a confirmation dialog and toast notifications when deletion success or fail.

image
image
image

While testing the feature, I identified a couple of issues we need to look into: 
1-Admins currently lack visibility into private templates, which conflicts with their privileges and the fact that they can access all user data.
2- there is no option to edit templates. **for me it's a design choice not really an issue but it's nice to have**
3- the recently used section is currently static showing the very first created template in the system and doesn't reflect actual history.

resolves #491


ashmod:
@unauthorised-401 will be reviewing the backend changes and I'll review the frontend changes shortly.
One tiny note, you mentioned you'd like to allow editing templates - but I think this sort of defeats the purpose of templates, they should be an immutable form of workflows for users to use, think: https://n8n.io/workflows/
If a user would like to modify a template, they are free to **save** it as a personal workflow, and use that instead.

With that in mind, I also think the history issue is important to address, in addition to the placeholder gmail image, those should definitely be looked into.

Lucifer3224:
I completely understand the point about keeping templates immutable. My reasoning was that if a creator builds a template and later needs to make a tiny adjustment, it's easier to edit it directly than to use, modify, re-save, and then delete the old version. Doing this feels more like a workaround than a clean solution. Also, even the template name is not editable, what if the user made a typo error? Should they delete it and create a new one? If it’s 2 or 3 nodes, okay, it’s easy but what if the workflow is complex? You mighr say they can export it, delete the old one, and create a new one with the correct name, okay again, but it’s still a workaround and it’s quite tedious. I specifically meant for **the creator** to have the ability to refine their own work directly without those extra steps. That’s why I still see it as a design choice to provide that flexibility.
As I said, of course, it’s not mandatory, I just wanted to clarify why I suggested that it's nice to have

ashmod:
Yes, creators are definitely allowed to edit the templates they publish - whether we allow this directly through the app or via other means (e.g. directly contributing to the repository.)

That said, some templates are going to be provided as-is from our root repository, and those would not be mutable in the sense that a creator would have control over them but rather they'd just live in the codebase and can be modified when needed. Templates are just JSONs after all.
We're also still studying the possibility of having templates live in a completely separate repo, so that would streamline how users would access/publish templates.

ashmod:
Frontend changes lgtm. @unauthorised-401 @Spafic if either of you has the time to give the api a quick scan before we merge this.

Spafic (dismissed):
@ashmod Api changes LGTM
The API changes look good. The int ID was replaced with a user object to allow access to additional attributes like 'Role' during deletion.

ashmod:
> allow the admin also to delete a template even if he is not the creator

Does this also include user-created non-public templates?

apps/web/src/app/(app)/create/templates/page.tsx
```diff
@@ -187,13 +254,46 @@ function TemplatesPageInner() {
...
+                        </AlertDialog>
+                      )}
+                      <Button
+                        variant="outline"
+                        size="sm"
+                        onClick={() => handleUseTemplate(template.id)}
+                      >
+                        Use
+                      </Button>
+                    </div>
```
ashmod: This looks identical to the block [above](https://github.com/rune-org/rune/pull/527/changes#diff-9af655940860d14fc6dfc41d97a86c36e66945d9fab9837b3270757b485ed32aR163), maybe we can extract to a shared, small component instead? Or if we could reuse one of our buttons that would be even better!

apps/web/src/app/(app)/create/templates/page.tsx
```diff
@@ -11,10 +11,25 @@ import { PageHeader } from "@/components/layout/PageHeader";
 import { listTemplates } from "@/lib/api/templates";
 import type { TemplateSummary } from "@/client/types.gen";
 import { useRouter, useSearchParams } from "next/navigation";
+import { useAuth } from "@/lib/auth";
+import { deleteTemplate } from "@/lib/api/templates";
```
ashmod: We can add this to the import above at L11

Lucifer3224:
> Does this also include user-created non-public templates?

nope, non-public template deletion is restricted to the creator only

ashmod:
> > Does this also include user-created non-public templates?
> 
> nope, non-public template deletion is restricted to the creator only

Well, I think we should revise our approach here then 🤔
Particularly because I noticed admins can not actually see user-created private templates either. Is this tracked somewhere else in another PR?

All in all, the admin role is basically the platform superuser, so they should not be restricted in any way.

Lucifer3224:
> > > Does this also include user-created non-public templates?
> > 
> > 
> > nope, non-public template deletion is restricted to the creator only
> 
> Well, I think we should revise our approach here then 🤔 Particularly because I noticed admins can not actually see user-created private templates either. Is this tracked somewhere else in another PR?
> 
> All in all, the admin role is basically the platform superuser, so they should not be restricted in any way.

I get that admins are superusers, but I intentionally excluded private templates from their scope. In my view, a private template is more like a user's personal draft or a private shortcut, just like admins don’t usually access a user’s unsent drafts or personal notes in other apps.

It's more like a safe space to experiment until users choose to make them public.

unauthorised-401:
> > > > Does this also include user-created non-public templates?
> > > 
> > > 
> > > nope, non-public template deletion is restricted to the creator only
> > 
> > 
> > Well, I think we should revise our approach here then 🤔 Particularly because I noticed admins can not actually see user-created private templates either. Is this tracked somewhere else in another PR?
> > All in all, the admin role is basically the platform superuser, so they should not be restricted in any way.
> 
> I get that admins are superusers, but I intentionally excluded private templates from their scope. In my view, a private template is more like a user's personal draft or a private shortcut, just like admins don’t usually access a user’s unsent drafts or personal notes in other apps.
> 
> It's more like a safe space to experiment until users choose to make them public.

I think restricting them here defeats the purpose of the admin being a superuser. Since admins already have the authority to manage sensitive resources like workflows and credentials, what makes templates the exception?

Regarding the already provided system templates, if the goal is to protect them, we can just differentiate them using a flag and completely restrict their deletion or editing

Lucifer3224:
> > > > > Does this also include user-created non-public templates?
> > > > 
> > > > 
> > > > nope, non-public template deletion is restricted to the creator only
> > > 
> > > 
> > > Well, I think we should revise our approach here then 🤔 Particularly because I noticed admins can not actually see user-created private templates either. Is this tracked somewhere else in another PR?
> > > All in all, the admin role is basically the platform superuser, so they should not be restricted in any way.
> > 
> > 
> > I get that admins are superusers, but I intentionally excluded private templates from their scope. In my view, a private template is more like a user's personal draft or a private shortcut, just like admins don’t usually access a user’s unsent drafts or personal notes in other apps.
> > It's more like a safe space to experiment until users choose to make them public.
> 
> I think restricting them here defeats the purpose of the admin being a superuser. Since admins already have the authority to manage sensitive resources like workflows and credentials, what makes templates the exception?
> 
> Regarding the already provided system templates, if the goal is to protect them, we can just differentiate them using a flag and completely restrict their deletion or editing

The flag approach makes perfect sense for protection, and I definitely suggest that.

As for the visibility, **I actually flagged that conflict in the PR description initially. But the more I looked at it, the more I started seeing these as a personal 'playground' for users.** 

I’m not looking at it from a security or data sensitivity perspective (since, as you said, admins already handle much more sensitive info). I just feel it's about the user's creative flow, a scratchpad where they can experiment and scribble freely before they're ready to go public. Since these templates aren't live and don't actually affect the platform's operations, I didn't see much functional value in having them visible to admins; I just thought keeping them private added a nice layer of user comfort.

unauthorised-401:
> > > > > > Does this also include user-created non-public templates?
> > > > > 
> > > > > 
> > > > > nope, non-public template deletion is restricted to the creator only
> > > > 
> > > > 
> > > > Well, I think we should revise our approach here then 🤔 Particularly because I noticed admins can not actually see user-created private templates either. Is this tracked somewhere else in another PR?
> > > > All in all, the admin role is basically the platform superuser, so they should not be restricted in any way.
> > > 
> > > 
> > > I get that admins are superusers, but I intentionally excluded private templates from their scope. In my view, a private template is more like a user's personal draft or a private shortcut, just like admins don’t usually access a user’s unsent drafts or personal notes in other apps.
> > > It's more like a safe space to experiment until users choose to make them public.
> > 
> > 
> > I think restricting them here defeats the purpose of the admin being a superuser. Since admins already have the authority to manage sensitive resources like workflows and credentials, what makes templates the exception?
> > Regarding the already provided system templates, if the goal is to protect them, we can just differentiate them using a flag and completely restrict their deletion or editing
> 
> The flag approach makes perfect sense for protection, and I definitely suggest that.
> 
> As for the visibility, **I actually flagged that conflict in the PR description initially. But the more I looked at it, the more I started seeing these as a personal 'playground' for users.**
> 
> I’m not looking at it from a security or data sensitivity perspective (since, as you said, admins already handle much more sensitive info). I just feel it's about the user's creative flow, a scratchpad where they can experiment and scribble freely before they're ready to go public. Since these templates aren't live and don't actually affect the platform's operations, I didn't see much functional value in having them visible to admins; I just thought keeping them private added a nice layer of user comfort.

That makes sense, and I don't mind taking the scratchpad approach. However, I think we really need to document the admin functionality going forward. The initial plan was for admins to have access to absolutely everything, so with these types of exceptions, it's getting a bit harder to keep track of what they actually can and can't see. Let's make sure we get the current admin scope documented @Lucifer3224 @ashmod

ashmod:
Is this good to merge?

ashmod:
Templates backend underwent some significant changes, it would be better to handle the new structure separately since this is outdated for the most part.

PR #545, opened 2026-04-21
fix: resolve stale UI state and execution failures for revoked credentials

### LINKED ISSUE #488: Deleted shared credential causes silent runtime failure
When User A deletes a shared credential while User B has a workflow open that uses it:
- User B's next run fails with no clear explanation.
- The credential still appears in the node inspector (stale cache).
- After refreshing, the credential silently disappears from node configs with no error message.

#### Steps to Reproduce

1. User A shares a credential with User B.
2. User B configures a workflow node using that credential and keeps the page open.
3. User A deletes the credential.
4. User B runs the workflow.

#### Expected Behavior

- The run failure should include a clear message: "Credential 'X' no longer exists or was revoked."
- The node inspector should indicate the credential is missing.

#### Actual Behavior

- Run fails with no useful error.
- Credential appears intact in the inspector until a page refresh.
- After refresh, the credential silently disappears from configs.

Lucifer3224:
Summary of what I've done:

Integrated real-time sync using **Redis Pub/Sub** and **SSE**
Created a global `useCredentialEvents` hook to observe changes
Added reactive **"(Missing)"** warnings and auto-refresh to the `CredentialSelector`

resolves #488


apps/web/src/hooks/useCredentialEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+      try {
+        const baseUrl = client.getConfig().baseUrl;
+        // Construct the full URL to the events endpoint
+        const url = `${baseUrl}/credentials/events`;
+
+        const { stream } = createSseClient({
+          url,
+          signal: abortController.signal,
+          // Need to include credentials to send the httpOnly access token cookie
+          credentials: "omit", // Wait, fetch API might need 'include' for cookies
```
ashmod: This is immediately overridden in `onRequest` with `init.credentials = "include"`. Can you elaborate on the reasoning here?

services/api/src/credentials/service.py
```diff
@@ -178,6 +181,13 @@ async def delete_credential(self, credential_id: int, user: User) -> None:
         await self.session.delete(credential)
         await self.session.commit()
 
+        # Publish event
+        redis = get_redis_client()
+        await redis.publish(
+            "credential_events",
+            json.dumps({"action": "deleted", "credential_id": credential_id}),
+        )
```
ashmod: This only publishes events for credential deletion, but the stale revoked-credential case is handled by the share/revoke endpoints rather than delete. In particular, [`revoke_credential_access`](https://github.com/rune-org/rune/blob/4ec29f4c0f99017932daa0ac3c938dcfb5c7f501/services/api/src/credentials/permissions.py#L179-L180) removes another user's access without publishing an event, so the SSE stream never notifies the affected user, and the (Missing) warning never appears for them.

services/api/src/credentials/router.py
```diff
@@ -111,6 +113,53 @@ async def list_credentials_dropdown(
...
+    Streams real-time events when credentials are deleted, shared, or revoked.
+    Useful for updating the UI instantly without polling.
+
+    Permissions:
+    - Any authenticated user can subscribe. They will only receive events
+      for credentials they have access to (handled by the client checking IDs).
+    """
+
+    async def event_generator():
+        from src.db.redis import get_redis_client
```
ashmod, line 137: This should go above with imports

apps/web/src/hooks/useCredentialEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// Create a custom event name
+export const CREDENTIAL_CHANGED_EVENT = "rune:credential_changed";
+
+/**
+ * Hook to listen to global credential events (deleted, updated, shared) via SSE.
+ * When an event is received, it dispatches a window event so components like
+ * CredentialSelector can react instantly.
+ */
+export function useCredentialEvents() {
```
ashmod, line 14: This hook is manually constructing the URL (${baseUrl}/credentials/events) and calling `createSseClient` directly instead of using the generated [`credentialEventsCredentialsEventsGet`](https://github.com/rune-org/rune/blob/4ec29f4c0f99017932daa0ac3c938dcfb5c7f501/apps/web/src/client/sdk.gen.ts#L812) function. This defeats the purpose of our client layer completely.

ashmod (commented):
~The API test ci seems to be failing as well, please review.~

apps/web/src/hooks/useCredentialEvents.ts
Lucifer3224: Honestly, while I was setting up the SSE connection, I was testing different credential settings and once I got it working, I guess I forgot to refactor that part back to its final version. I'm sorry for this, I've just updated the code to set it to include directly.

services/api/src/credentials/service.py
Lucifer3224: Yeah because I was primarily focused on the delete flow as the main event and didn't initially account for the revoke needing the same notification logic, anyway I've refactored the service to handle both cases correctly.

services/api/src/credentials/router.py
Lucifer3224: I put it here to keep it lazy since this is a long-lived connection, I wanted it to be loaded when a user hits this endpoint.

apps/web/src/hooks/useCredentialEvents.ts
Lucifer3224: I didn't realize the generated client had `.sse` namespace, so I fell back to using the core `createSseClient` helper. Thanks for pointing that, I've now refactored it to use `client.sse.get`

ashmod (approved):

PR #562, opened 2026-04-22
feat: implement credential dependency safety checks and usage tracking

### LINKED ISSUE #487: Deleting a credential referenced in a workflow gives no warning
Deleting a credential that is actively referenced in one or more workflows succeeds silently with no dependency check or warning.

#### Steps to Reproduce

1. Create a credential and use it in a workflow node.
2. Go to the credentials page and delete that credential.

#### Expected Behavior

A warning dialog listing the workflows that depend on this credential, with a confirm/cancel option.

#### Actual Behavior

The credential is deleted immediately with no warning.

Lucifer3224:
Integrated backend tracking using a **WorkflowCredentialLink** table to index active references
Exposed a new `GET /usage` endpoint to retrieve dependent workflows and their owners
Implemented a polished `DeleteCredentialDialog` with a scrollable dependency list and owner information
image

resolves #487


ashmod:
Build is failing, my guess is the last merge handled the client regen poorly. Try regenerating and pushing the changes.

services/api/src/credentials/service.py
```diff
@@ -178,6 +185,31 @@ async def delete_credential(self, credential_id: int, user: User) -> None:
...
+
+        Args:
+            credential_id: ID of the credential
+
+        Returns:
+            List of (Workflow, User) tuples referencing this credential
+        """
+        statement = (
+            select(Workflow, User)
+            .join(WorkflowCredentialLink)
```
ashmod, line 245: Existing workflows that already reference credentials will have no link rows until someone edits them manually, so `/usage` can incorrectly return no dependencies immediately after deploy.

That makes the delete warning unreliable for upgraded environments. Please add a backfill migration/job, or derive usage from existing workflow version JSON until the link table is fully populated.

apps/web/src/lib/api/setupClientInterceptors.ts
```diff
@@ -96,10 +96,13 @@ export function setupClientInterceptors() {
...
+    headers.delete("Authorization");
     const result = await client.request({
       ...options,
+      security: [],
       headers,
       method: (options.method ?? "GET") as Uppercase<HttpMethod>,
+      parseAs: "stream",
     });
     return result.response;
   });
```
ashmod, line 111: The changes in this file don't seem relevant to this PR. Could you elaborate on why those are added here?

services/api/src/workflow/service.py
```diff
@@ -187,6 +188,10 @@ async def create_version(
 
         locked_workflow.latest_version_id = version.id
         locked_workflow.is_active = locked_workflow.published_version_id is not None
```
ashmod, line 230: This syncs credential links from the latest saved version, but deletion safety needs to reflect the workflow version that can actually run. If v1 is published with credential C, then an unpublished v2 removes C, this call removes the link even though active runs still depend on C. `/usage` can then report no dependencies and allow deleting C, breaking the published workflow.

Consider syncing links from the published version, or tracking draft and published usage separately and using published usage for delete safety.

services/api/src/workflow/service.py
```diff
@@ -231,6 +236,10 @@ async def restore_version(
...
```
ashmod, line 283: Same issue here: restoring a version updates the workflow-level credential links based on the latest/restored draft state, not necessarily the currently published version. If the restored version is not published, this can make `/usage` diverge from the workflow version that active runs execute.

For delete safety, the link table should represent published dependencies, or the API should distinguish draft usage from active published usage.

services/api/src/credentials/router.py
```diff
@@ -201,6 +202,34 @@ async def delete_credential(
...
+    credential_id: int,
+    current_user: User = Depends(require_password_changed),
+    service: CredentialService = Depends(get_credential_service),
+) -> ApiResponse[list[CredentialUsage]]:
+    """
+    Get a list of workflows that are currently using this credential.
+
+    """
+    # Verify user has access to the credential first
+    await service.get_credential(credential_id, current_user)
```
ashmod:
```suggestion
    credential = await service.get_credential(credential_id, current_user)
    await service.permission_service.require_delete_access(credential.data, current_user)
```

apps/web/src/components/credentials/DeleteCredentialDialog.tsx
```diff
@@ -1,33 +1,102 @@
...
+      console.error(message, err);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
   return (
     <ConfirmationDialog
       open={open}
       onOpenChange={onOpenChange}
```
ashmod, line 62:
```suggestion
      onOpenChange={onOpenChange}
      isLoading={isLoading}
```

apps/web/src/components/credentials/DeleteCredentialDialog.tsx
```diff
@@ -1,33 +1,102 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    if (open && credentialId) {
+      loadUsage();
+    } else {
+      setUsage([]);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [open, credentialId]);
```
ashmod: This effect can apply stale usage results if the dialog closes/reopens or `credentialId` changes while `loadUsage()` is still in flight. The old request can resolve later and populate the dialog with dependencies for the previous credential.

Can we inline the fetch in the effect with a cancellation guard, or otherwise ignore results when the effect has been cleaned up? That would also remove the need to suppress `react-hooks/exhaustive-deps`.

For example, the effect could set a `cancelled` flag in cleanup and only call `setUsage` / `setIsLoading(false)` when it is still current.

apps/web/src/components/ui/confirmation-dialog.tsx
```diff
@@ -48,7 +48,9 @@ export function ConfirmationDialog({
       <DialogContent>
         <DialogHeader>
           <DialogTitle>{title}</DialogTitle>
-          <DialogDescription>{description}</DialogDescription>
+          <DialogDescription asChild>
+            <div>{description}</div>
+          </DialogDescription>
```
ashmod, line 53: This changes the semantic element from `<p>` to`<div>`. This is functionally fine but may slightly affect screen readers. The real fix is accepting ReactNode in the description prop type (which it already does), so the wrapper is not needed. Consider removing the `asChild` and just accepting that `<p>` will contain block-level children via the existing description prop pass-through.

services/api/src/workflow/service.py
```diff
@@ -730,3 +739,60 @@ def _normalize_message(message: str | None) -> str | None:
...
+            valid_stmt = select(WorkflowCredential.id).where(
+                col(WorkflowCredential.id).in_(list(credential_ids))
+            )
+            valid_res = await self.db.exec(valid_stmt)
+            valid_credential_ids = set(valid_res.all())
+
+            # Find existing link IDs for this workflow to avoid duplicates
+            existing_stmt = select(WorkflowCredentialLink.credential_id).where(
+                WorkflowCredentialLink.workflow_id == workflow_id
+            )
```
ashmod: non-blocking, but I think the SELECTs here can be joined, right? This would save one round-trip per version save.

apps/web/src/lib/api/setupClientInterceptors.ts
Lucifer3224: because after the client generation update, the interceptor logic was throwing compilation errors and they were blocking the build, so I had to resolve them to continue.

apps/web/src/components/ui/confirmation-dialog.tsx
Lucifer3224: Actually, leaving it as a `<p>` tag will cause a hydration error. I’ve run into this before, so I kept it as is.

Lucifer3224:
Could you check the updates and let me know if you have any further comments, @ashmod

apps/web/src/lib/api/setupClientInterceptors.ts
ashmod: Thanks for clarifying. The `headers.delete("Authorization")` and `security: []` changes probably make sense if the retry needs to avoid reusing stale auth data after refresh.

For `parseAs: "stream"`, is the intent to prevent the inner retry request from consuming the response body before returning `result.response` to the outer request? If so, I guess that makes sense, but it would be useful to document or test this because it is not obvious from the code and this interceptor affects every retried API request.

Could you also share the original compilation error from the client generation update? I want to confirm this is the minimal type-safe fix rather than an unrelated behavior change.

Lucifer3224: sorry, I didn't notice this comment I was checking why the pr is not closed yet and just noticed it, so yup, I added `parseAs: "stream"` to ensure the body remains untouched so the calling site can handle the parsing as originally intended because without it caused this error for the original caller: `TypeError: Failed to execute 'json' on 'Response': body stream already read`

As for the document, would a detailed code comment explaining the 'body already read' issue be enough, or do you mean somewhere else?

ashmod: Thanks for clarifying! Yes a small comment above the parseAs with reasoning on why we went with "stream" should be enough.

Lucifer3224: done

ashmod (approved):

PR #564, opened 2026-04-24
feat(web): align HTTP canvas node with DSL parameters

### LINKED ISSUE #563: [Improvement]: align HTTP canvas node with DSL parameters
### Improvement Summary

Align the canvas **HTTP** node’s TypeScript model (`apps/web/src/features/canvas/types.ts`) with the full **`http`** parameter set from `dsl/dsl-definition.json`, using **camelCase** property names in TypeScript. DSL/worker **snake_case** keys remain the persistence/runtime contract; the frontend owns a clear camelCase shape (and any serialize/deserialize mapping at the API or workflow boundary).

### Current Behavior

- `NodeDataMap["http"]` only covers a subset of fields in camelCase (`method`, `url`, `headers`, `query`, `body`, `timeout`, `retries`, `ignoreSSL`).
- `method` omits **`PATCH`** relative to the DSL enum.
- **`raiseOnStatus`** and **`retryDelay`** are not modeled.
- **`retries`** (number) does not match the DSL/worker concept of **`retry`** (string count) and naming.
- **`timeout`** is a number in TS while the DSL describes seconds as strings (worker accepts several coercions); types do not document the intended wire shape.
- Overall, TS types under-represent the DSL and risk drifting from what executions actually support.

### Proposed Improvement

- Expand the **`http`** branch of `NodeDataMap` (or a dedicated `HttpNodeData` type) so it covers every DSL `nodes.http.parameters` key, **expressed in camelCase**, for example:
  - `retry`, `retryDelay`, `timeout`, `raiseOnStatus`, `ignoreSsl` (or `ignoreSSL` if you standardize that token—pick one and use it consistently).
- Include **`PATCH`** in the `method` union.
- Choose TypeScript types that match **authoring** (e.g. `number` for timeouts in the UI) **or** match **wire** (`string`) explicitly; if they differ, document a single mapper to snake_case for save/load.
- Centralize **camelCase ↔ snake_case** mapping in one place (e.g. workflow serialization or API client) so node components only use camelCase.

### Benefits

- **Idiomatic TypeScript** while keeping worker/DSL snake_case at the boundary.
- **Full surface area** for the HTTP node in types and autocomplete, aligned with `dsl-definition.json`.
- **Explicit mapping** reduces bugs from ad hoc field renames (`retries` vs `retry`, `ignoreSSL` vs `ignore_ssl`).
- **Easier reviews**: new DSL parameters become a checklist for TS + mapper + UI.

### Additional Context

_No response_

### Pre-submission checklist

DopeBiscuit:
## Summary

Aligns the canvas **HTTP** node with the DSL/worker: full parameters, **snake_case** `data` fields matching saved `parameters`, **`PATCH`**, and **`workflow-dsl`** encode/decode (with legacy **`retries`** / old camel keys).

## Changes

- Types: `HTTP_METHODS` / `HttpMethod`, `retry`, `retry_delay`, `raise_on_status`, `ignore_ssl`
- Inspector: advanced fields + method select without cast; short raise-on-status hint
- `nodeRegistry`: defaults for new HTTP nodes (`4xx,5xx`, retry/delay/timeout)
- Tests: HTTP round-trip + legacy `retries`

Closes #563


ashmod:
Typo: update linked issue to 563

ashmod (approved):
Changes look good!

Non-blocking: We would like users to configure status-code success explicitly (overriding the default halt/failure)

You can decide on whether this should just be a frontend change, or if you'd like to add a separate function e.g. `ignore_on_status` on the worker level (which I prefer).

I think it would be best to handle this in a follow-up, wdyt?

DopeBiscuit:
If it requires work in the worker then yea a follow-up would be better, but can you elaborate further ?

ashmod:
> If it requires work in the worker then yea a follow-up would be better, but can you elaborate further ?

Basically we just want to allow users to configure, through the HTTP inspector, if they would like to treat error codes as errors (so for example the default 4xx and 5xx would halt as they currently do). Or, for whatever reason, they'd like them to act as success i.e. not halt the workflow. One use case for that would be a user checking if a certain page 404s and act on it (e.g. via conditional check.) Does that make more sense?

PR #566, opened 2026-04-24
feat(web): align frontend conversion logic with new DSL schema

Lucifer3224:
Updates the frontend's workflow conversion and data handling logic to match the updated DSL schema, resolving persistent validation mismatches during workflow saves


ashmod:
From my understanding, this PR should be syncing the `dsl/` generated types with the frontend's dsl. But what's actually happening is a local copy of the file, so they can easily drift and cause the same problem.

But I notice there's a `@dsl/*` alias pointing to the `dsl/generated/` directory which should be the correct approach - but it's never used. 

Could you elaborate on this?

dsl/dsl-definition.json
```diff
@@ -118,24 +136,35 @@
...
+            "header",
+            "token",
+            "custom",
+            "smtp"
+          ]
         },
         "values": {
           "type": "object",
-          "required": true,
+          "required": false,
```
ashmod, line 151: Is this intentional?

apps/web/src/features/canvas/lib/executionConversion.ts
```diff
@@ -42,8 +42,10 @@ export function extractGraphSnapshot(
       trigger,
       parameters: (extra.parameters as Record<string, unknown>) ?? {},
       output: (extra.output as Record<string, unknown>) ?? {},
+      error: undefined,
+      credential_type: undefined,
```
ashmod, line 46: Not required since these are optional fields in the type, and TypeScript doesn't require them to be explicitly set to undefined.

apps/web/src/lib/workflow-dsl.test.ts
```diff
@@ -81,6 +83,8 @@ describe("workflow DSL helpers", () => {
           type: "ScheduledTrigger",
           parameters: { amount: 10, unit: "minutes" },
           output: {},
+          error: undefined,
+          credential_type: undefined,
```
ashmod, line 87: Not required here as well, and across the rest of this file.

dsl/dsl-definition.json
```diff
@@ -145,32 +174,55 @@
...
-
     "ScheduledTrigger": {
       "description": "Scheduled workflow trigger with interval-based execution",
       "credential_type": null,
       "parameters": {
-        "amount": { "type": "number", "required": true, "description": "Quantity of time between executions" },
-        "unit": { "type": "string", "required": true, "enum": ["seconds", "minutes", "hours", "days"], "description": "Unit of time for the interval" }
+        "amount": {
+          "type": "number",
+          "required": false,
```
ashmod, line 189: why?

dsl/dsl-definition.json
```diff
@@ -145,32 +174,55 @@
...
+          "description": "Quantity of time between executions"
+        },
+        "unit": {
+          "type": "string",
+          "required": false,
```
ashmod, line 194: also why here too?

apps/web/src/lib/workflow_dsl_generated.ts
```diff
@@ -0,0 +1,951 @@
...
+  id: string;  // Unique identifier for the node within the workflow
+  name: string;  // Human-readable node name
+  trigger: boolean;  // Whether this node initiates workflow execution
+  output: Record<string, any>;  // Placeholder for execution output (empty in definition)
+  error: ErrorHandling | undefined;  // Error handling configuration
+  credential_type: string[] | undefined;  // List of allowed credential types for this node (for UI filtering)
+  credentials: Credential | undefined;  // Complete credential object with values
+}
+
+export function sanitizeBaseNode(obj: BaseNode): { valid: boolean; errors: string[] } {
```
ashmod, line 769: are the sanitize* functions from this file used anywhere?

PR #575, opened 2026-04-27
test(api): Add comprehensive API level tests for workflow endpoints

ZeynatGhallab:
Added (4 New Test Files):
1- test_workflow_permissions.py — VIEWER/EDITOR/OWNER/ADMIN role enforcement 
2- test_workflow_execution_and_state.py — Execution lifecycle, publishing, state tracking 
3- test_workflow_input_validation.py — Input validation, error codes 
4- test_workflow_admin_and_failures.py — Admin bypass, timeouts, partial failures

Improved (3 Refactored Files)
1- test_workflow_api.py — Removed redundant tests; kept queue/conflict verification 
2- test_workflow_persistence.py — Converted from service-layer to API-level assertions 
3- test_workflow_edge_cases.py — Enhanced with meaningful edge case scenarios


ashmod:
@ZeynatGhallab Good PR overall, the test infrastructure work is solid and was needed. The shared-connection fix, the cookie-isolation fix for multi-user tests, and the schema validation additions are also nice improvements.

That said, running the suite surfaced 8 failures that need to be addressed before this merges.

3 are schema validation gaps in `_validate_workflow_data`:
- Self-referencing edges (src == dst) are accepted
- Edges whose src or dst don't reference any node in the same workflow are accepted

5 are the sharing write endpoints being broken:
- `POST /workflows/{id}/share`
- `DELETE /workflows/{id}/share/{user_id}`
- `PATCH /workflows/{id}/permissions/{user_id}`

`GET` works, so it's isolated to the write paths. Run:
```py
cd services/api && uv run pytest test/workflow/test_workflow_permissions.py::TestWorkflowSharing -v
``` 
to see the exact responses.

A few things that are fine but worth noting for follow-up:

The concurrent version creation test runs sequentially by necessity - the test setup shares a single DB connection across all HTTP sessions, so `asyncio.gather` on writes causes an `InterfaceError`. The conflict detection logic is still tested correctly, but the actual DB-level race (two transactions simultaneously reading the same `latest_version_id`) isn't covered. That would require a separate connection pool setup in tests. Not a blocker, but worth tracking.

The SAML endpoints and the internal run endpoint still have zero test coverage. Also should we add tests now for Smith/Scryb? cc @unauthorised-401 . Those are probably not expected in this PR, but should be tracked somewhere so it doesn't get forgotten.

Overall the test suite is meaningfully better than before. cc @Spafic  and @unauthorised-401  for thoughts on this, we should ideally get those 8 tests to pass before merging.

Spafic (dismissed):
LGTM. One quick tweak: our DSL generator produces a Python file that serves as the core of our validations. This solution is temporary until PR #565 is implemented.

Why do we need DSL-based validations? 
Consider users attempting to inject undefined nodes into our workflow.

Spafic:
- test_workflow_with_self_referencing_edge_rejected
- test_edge_referencing_nonexistent_src_node_rejected
- test_edge_referencing_nonexistent_dst_node_rejected 


The following tests have a permission typo: our backend accepts lowercase role strings (VIEWER = "viewer", EDITOR = "editor", OWNER = "owner"), while the tests use capitalized versions as strings instead of objects.

- [False] test_owner_can_share_workflow_with_viewer_role 
- [False] test_owner_can_share_workflow_with_editor_role
- [False] test_share_with_nonexistent_user_returns_error
- [False] test_owner_can_revoke_access
- [False] test_owner_can_update_user_role

The first three tests raise valid concerns. Currently, the API doesn't validate data in WFS. I recommend marking these as P3, since the frontend prevents self-referencing under normal conditions, so this shouldn't be a priority.

Spafic:
Depends on #631

ashmod (approved):

PR #584, opened 2026-04-29
feat(mcp): add mcp client in go worker

### LINKED ISSUE #583: Implement MCP client
### Feature Summary

Implement MCP client interface

### Problem / Use Case

Accelerate integrations

### Proposed Solution

_No response_

### Additional Context

_No response_

### Pre-submission checklist

Spafic:
Implemented client with connect, discover, and call utilities, along with required interface and mock files for integration.

More details in Readme.

- closes #583


services/rune-worker/pkg/mcp/integrations/microsoft/outlook/outlook.go
```diff
@@ -0,0 +1,25 @@
...
+import "rune-worker/pkg/mcp"
+
+func init() {
+	mcp.RegisterIntegration(mcp.IntegrationConfig{
+		Name: "outlook",
+		URL:  "http://outlook-mcp:3300/mcp",
+		Tools: []mcp.ToolDef{
+			{
+				MCPName:     "send_email",
+				NodeName:    "send_email",
```
DopeBiscuit: Why is this different from the rest ?

DopeBiscuit:
You didn't test out the tools yet, right ? 

Because you are missing oauth.

Spafic:
> You didn't test out the tools yet, right ?
> 
> Because you are missing oauth.

yes just mock

DopeBiscuit:
Ok, I am currenlty reviewing and testing the code.

services/rune-worker/pkg/mcp/node.go
```diff
@@ -0,0 +1,75 @@
...
+type MCPNode struct {
+	manager      *Manager
+	providerName string
+	toolName     string
+	arguments    map[string]any
+	timeout      time.Duration
+}
+
+// NewMCPNode builds an MCPNode from the execution context.
+func NewMCPNode(manager *Manager, provider, tool string, execCtx plugin.ExecutionContext) *MCPNode {
```
DopeBiscuit, line 23: Why do you pass the provider and tool instead of pushing them to execCtx.params ?

services/rune-worker/pkg/mcp/manager.go
```diff
@@ -0,0 +1,100 @@
...
+		p.Disconnect()
+	}
+}
+
+// AddProvider adds a provider directly (tests only).
+func (m *Manager) AddProvider(p *Provider) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.providers[p.Name()] = p
+}
```
DopeBiscuit: Why do you not use any of these helper functions ? Excluding the DisconnectAll() this you use I know.

But why do you not use the rest instead of manually repeating them again.

services/rune-worker/pkg/mcp/provider.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil, fmt.Errorf("provider %s: not connected", p.name)
+	}
+
+	var tools []*mcp.Tool
+	for tool, err := range session.Tools(ctx, nil) {
+		if err != nil {
+			return nil, fmt.Errorf("list tools from %s: %w", p.name, err)
+		}
+		tools = append(tools, tool)
+	}
```
DopeBiscuit, line 80: What is the purpose of this function ? It is only used in the bridge_test, but according to you readme, our only tools should be well defined on startup not on execution time, so that the dsl knows all the tools that we have, so we don't need to discover server tools since we run well-defined nodes.

services/rune-worker/pkg/mcp/provider.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Disconnect closes the MCP session.
+func (p *Provider) Disconnect() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.closed || p.session == nil {
+		return
+	}
+	p.closed = true
```
DopeBiscuit: This variable is only changed in this function and changed to true, it never gets set back to false.

This means that if you open the reconnection for a certain provider you cannot close it again, only wait for it to expire.

DopeBiscuit (commented):
Add example arguments

DopeBiscuit:
Sebak mn el comments, I will resolve this, extension gone crazy.

DopeBiscuit (dismissed):

services/rune-worker/pkg/mcp/manager.go
```diff
@@ -0,0 +1,108 @@
...
+
+	cfg, ok := m.configs[name]
+	if !ok {
+		return nil, fmt.Errorf("mcp integration %q not configured", name)
+	}
+
+	p := NewProvider(name)
+	if err := p.ConnectHTTP(ctx, cfg.URL); err != nil {
+		return nil, fmt.Errorf("mcp connect %s: %w", name, err)
+	}
```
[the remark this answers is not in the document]
DopeBiscuit: kont fen men f el 2 reviews el fato ya3ni ya 7elw enta

DopeBiscuit (approved):
gemini review points have been turned into issue, they don't block the current functionality, they are important to refactor, but currently we can do without them for v0.

PR #592, opened 2026-05-01
feat(frontend): add node integrations ui

ashmod:
This PR initializes the frontend work for third-party integrations. Backend wiring is required to get the nodes to actually work so I'm keeping this as draft while we implement the worker-level logic.


apps/web/src/features/canvas/components/inspectors/IntegrationInspector.tsx
```diff
@@ -0,0 +1,213 @@
...
+            placeholder={field.placeholder}
+            onChange={(event) => {
+              const nextValue = event.target.value;
+              updateArgument(field.name, nextValue === "" ? undefined : Number(nextValue));
+            }}
+          />
+          {field.help && <div className="mt-1 text-xs text-muted-foreground/70">{field.help}</div>}
+        </div>
+      );
+    }
```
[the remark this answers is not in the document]
ashmod: These are mock/placeholder fields for a backend that doesn't exist yet. The actual argument fields and their types will be defined by whatever the backend implements. The number type in the catalog is just a UI hint we put in for now.

apps/web/src/features/canvas/integrations/catalog.ts
```diff
@@ -0,0 +1,322 @@
...
+        type: "select",
+        options: VALUE_INPUT_OPTIONS,
+      },
+      {
+        name: "values",
+        label: "Values",
+        type: "json",
+        required: true,
+        help: 'A two-dimensional array, for example [["Name", "Status"]].',
+      },
```
DopeBiscuit, line 171: Did you mean to make the type:"json" ?  I have no clue what the possible types are, is there an array type ? 


Image

Just asking because of like how this is configured, I assume the "advanced JSON" gets created in the inspector since it is not here in this catalog file, I have yet to review it, just asking for clarification.

What I really care about is that in the end the data would be sent to me in the backend similar to the "advanced JSON" filed, correct ?

ashmod: Yep, `type:"json"` is intentional because there is no `array` type in our inspector, the types we have currently are:
`text`,`textarea`,`number`,`boolean`,`select`, and `json`.

The advanced json panel is auto-generated by the inspector, it mirrors all argument fields as a single editable json object for those who prefer to write in json for whatever reason. The Values editor and the Advanced json both [write into `node.data.arguments`](https://github.com/rune-org/rune/pull/592/changes#diff-128f983104074dbe0c1ebe4fd56e96e813717be6f6e5c44310f079a3224bc031R203-R207), so the backend receives the same payload shape either way. So for this node you'd receive `{ "value_inpسut_option": "USER_ENTERED", "values": [[]] }`.

DopeBiscuit (approved):

DopeBiscuit (approved):
Clean stuff.

DopeBiscuit:
Something that I have missed, a few PRs recently added extra nodes / node parameters and we have not updated the dsl-definition that we use under the `dsl/` directory, this is something that needs to be done in the near future so that we don't get overrun by the amount of changes

PR #593, opened 2026-05-01
feat(api): add webhook trigger for workflow execution

seif-elwarwary:
Adds POST /webhook/{guid} — an unauthenticated endpoint that lets external systems fire a workflow just like clicking Run, with the POST body passed under "$trigger" in accumulated_context.

Changes:
- Add WebhookRegistration model + Alembic migration
- Register/deactivate webhook in workflow lifecycle (activate, deactivate, publish, bulk-set-status); GUID is generated once and never regenerated
- Fix workflow delete to explicitly remove WebhookRegistration rows before deleting the parent, matching the existing ScheduledWorkflow pattern
- Extend WorkflowQueueService.publish_workflow_run with accumulated_context
- Add POST /webhook/{guid} router + service (src/webhook/)
- Wire webhook router into app.py
- Unit tests: service-level lifecycle (7 cases) + API integration (7 cases)
- E2E test suite in test/e2e/ using a real uvicorn subprocess against Docker test containers (postgres:5433, redis:6380, rabbitmq:5673)


ashmod:
Before we can work on this on the frontend I think we need the dsl changes under `dsl/` to be in place, and the API schema updates in `services/api/src/workflow/schemas.py` for the client regen.

DopeBiscuit:
> Before we can work on this on the frontend I think we need the dsl changes under `dsl/` to be in place, and the API schema updates in `services/api/src/workflow/schemas.py` for the client regen.

Can you elaborate ? If you are talking about importing types/nodes from the generated dsl then we need to discuss this asap.

ashmod:
> > Before we can work on this on the frontend I think we need the dsl changes under `dsl/` to be in place, and the API schema updates in `services/api/src/workflow/schemas.py` for the client regen.
> 
> Can you elaborate ? If you are talking about importing types/nodes from the generated dsl then we need to discuss this asap.

yes the part where the dsl needs to be updated to reflect the new trigger type.

PR #617, opened 2026-05-12
feat(workflows): add API and UI support for workflow description updates

SeifT101:
This pull request adds support for editing workflow descriptions in the web application. The main changes include new API types and endpoints for updating workflow descriptions, as well as UI updates to allow users to edit and submit new descriptions for workflows directly from the workflows table.

Notes:
- Descriptions currently don't have any validation, i think only validation needed would for the maximum length, lmk how to proceed.
- In the UI, there's this question mark indicator that shows the full description in case of long or multi-lined description, lmk if this is the fine or if it should be changed
image


services/api/src/workflow/router.py
```diff
@@ -270,6 +271,27 @@ async def update_name(
     return ApiResponse(success=True, message="Workflow name updated", data=detail)
 
 
+@router.put("/{workflow_id}/description", response_model=ApiResponse[WorkflowDetail])
```
Spafic, line 274: You should use PATCH insted of PUT.
PUT replaces the entire resource, while PATCH applies only the changes needed to update it.

Spafic: We need to refactor as well the other put endpoints for status, name later

Spafic (changes requested):

ashmod:
@SeifT101 please address the review comments

PR #631, opened 2026-05-13
feat(validation): implement modular validation system for workflow data

### LINKED ISSUE #630: [Improvement]: Implement validation layer for workflows
### Improvement Summary

Currently, the wfs is missing DSL edge validations, which are required to implement schema validations.

### Current Behavior

WFS saved successfully but failed during execution.

### Proposed Improvement

validation layer on the schemas

### Benefits

_No response_

### Additional Context

_No response_

### Pre-submission checklist

Spafic:
- Closes #630

Changes:
- Added src/workflow/validation/ directory with modular validators:
  - base.py: ValidationError, ValidationResult, Validator interfaces
  - node.py: Node ID validation (missing ID, duplicates, type)
  - edge.py: Edge wiring validation (src/dst references, self-references)  
  - workflow.py: Combined validators
- Updated workflow/schemas.py to use validate_workflow_data() via 
  model_validator for strict workflow validation
- Updated workflow/queue.py to keep runtime validation separate from 
  schema validation (different concerns)
- Removed duplicate validation from templates/service.py


services/api/src/workflow/validation/node.py
```diff
@@ -0,0 +1,134 @@
...
+        seen_ids: set[str],
+    ) -> ValidationError | None:
+        """Validate a single node."""
+        if not isinstance(node, dict):
+            return NodeNotObjectError(index)
+
+        if "id" not in node:
+            return NodeMissingIdError(index)
+
+        node_id = node.get("id")
```
ashmod: This should also reject non-string IDs before duplicate tracking. Right now any truthy non-string value, such as `123`, passes this validator even though the docstring above says node IDs must be strings and the rest of the workflow code treats IDs as string references.

Consider adding an explicit type check here, e.g. return a validation error when `not isinstance(node_id, str)`.

services/api/src/workflow/validation/edge.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def _validate_edges(
+        self,
+        edges: list[Any],
+        node_ids: set[str],
+    ) -> list[ValidationError]:
+        """Validate all edges reference valid nodes."""
+        errors: list[ValidationError] = []
+
+        for idx, edge in enumerate(edges):
```
ashmod: This silently accepts malformed edge entries. Since `WorkflowStructureValidator` only verifies that `edges` is a list, payloads like `edges: [null]` or `edges: ["bad"]` pass validation instead of being rejected.

This branch should append a validation error for non-object edges rather than `continue`, otherwise invalid workflow data can still be saved.

services/api/src/workflow/validation/workflow.py
```diff
@@ -0,0 +1,205 @@
...
+
+    Runs:
+    1. Structure validation (basic shape)
+    2. Node ID validation (node IDs, uniqueness)
+    3. Edge wiring validation (edge connections)
+    """
+
+    def __init__(self):
+        super().__init__(
+            [
```
ashmod: `WorkflowValidator` currently runs node and edge validators even when structure validation fails. That can produce noisy or misleading follow-up errors for malformed input, for example when `nodes` or `edges` is missing or has the wrong type.

Since structure validation is the gate for the expected shape, consider short-circuiting here: run `WorkflowStructureValidator` first, and only run node/edge validators if it succeeds.

services/api/src/workflow/validation/edge.py
```diff
@@ -0,0 +1,272 @@
...
+    Returns:
+        List of validation errors
+    """
+    validator = EdgeWiringValidator()
+    workflow_data = {"nodes": [], "edges": edges}
+    result = validator.validate(workflow_data)
+    return result.errors
+
+
+def validate_no_self_referencing_edges(
```
ashmod: This duplicates the self-reference check already implemented in [`EdgeWiringValidator._validate_edges()`](https://github.com/rune-org/rune/pull/631/changes#diff-fa8d8249f15a3543a40b8829ae1f0ba44f9180957d185ddd98c8a6ab3cb37de7R153), and I don't see callers that need this separate path. Keeping duplicate validation logic makes it easier for behavior to drift.

Consider removing this helper, or routing it through the main edge validator so there is one source of truth.

services/api/src/workflow/validation/node.py
```diff
@@ -0,0 +1,134 @@
...
+        for i, node in enumerate(nodes):
+            node_error = self._validate_node(node, i, seen_ids)
+            if node_error:
+                errors.append(node_error)
+
+        if errors:
+            return ValidationResult.failure(errors)
+        return ValidationResult.success()
+
+    def _validate_node(
```
ashmod: This returns only the first error for a node, so a node missing both `id` and `type` only reports the missing ID. That behavior can be fine, but it is inconsistent with the broader `CompositeValidator` design that collects validation errors.

Consider either collecting all node-level errors or documenting that node validation intentionally reports the first blocking error per node.

services/api/src/workflow/validation/workflow.py
```diff
@@ -0,0 +1,205 @@
...
+    def validate(self, data: dict[str, Any]) -> ValidationResult:
+        """Validate workflow structure.
+
+        Args:
+            data: Workflow data to validate
+
+        Returns:
+            ValidationResult with any errors found
+        """
+        if not isinstance(data, dict):
```
ashmod: Similarly, this validator early-returns on the first structural error, while `CompositeValidator` collects errors from all validators. The mixed behavior makes the validation result shape harder to reason about.

If first-error behavior is intentional for structural validation, a short comment would help. Otherwise, consider collecting structural errors consistently.

services/api/src/workflow/schemas.py
```diff
@@ -167,6 +168,15 @@ class WorkflowCreateVersion(BaseModel):
     )
     message: Optional[str] = Field(default=None, description="Revision message")
 
+    @model_validator(mode="after")
+    def _validate_workflow_data(self) -> "WorkflowCreateVersion":
```
ashmod, line 248: This collapses structured validation errors into one string, which drops the `field` and `context` data carried by each `ValidationError`.  I think that would make it harder for API consumers or the UI to identify which node/edge/field failed.

Consider preserving the structured errors in the response, or at least formatting them in a way that keeps the field path available for consuming later.

services/api/src/workflow/validation/base.py
```diff
@@ -0,0 +1,117 @@
+"""Base validation classes and interfaces.
+
+Provides:
+- ValidationError: Base exception for validation failures
+- ValidationResult: Result object for validation operations
+- Validator: Abstract base class for validators
```
ashmod: Let's remove those, they are already covered in the class docstrings.

ashmod (approved):

PR #640, opened 2026-05-24
feat(templates): rebuild template gallery and integrate rune-templates submodule

ashmod:
Splits curated workflows into the [rune-templates](https://github.com/rune-org/rune-templates) repo (consumed here as a submodule) and rebuilds `/create/templates` into a scope-aware gallery with official/community/personal buckets, scope-filtered category counts, and visually distinct cards. 
 
  ## Summary

  - Rebuilds the templates feature end-to-end: curated content moves to a separate `rune-templates` repo (linked here as a git submodule at `services/api/templates_bundle/`) so community contribution is a low-friction GitHub PR.
  - Adds a migration for the new rune templates schema and structure
  - New gallery at `/create/templates` with All / Official / Community / Personal tabs, scope-aware category chips, debounced search, sortable cards, and a category-tinted preview banner with node-count chip.
  - New canvas action **Submit to global gallery** generates a ready-to-PR bundle JSON (credentials + runtime noise stripped) that contributors drop into `rune-templates/templates/<category>/`.
  - Backend hardens the template surface: strict `TemplateCategory` enum, typed `WorkflowGraph` Pydantic schema, derived `scope` field, idempotent bundle seeder with orphan removal, and a CI-checked JSON Schema export to keep both repos in sync.
  - `SaveTemplateDialog` refresh: Personal / Community visibility chooser, tags chip input, category selector backed by the enum.

  ## What's in rune vs rune-templates

  | Concern | Lives in |
  |---|---|
  | Workflow schema (`WorkflowGraph`, `TemplateCategory`) | rune |
  | JSON Schema (auto-exported from Pydantic) | rune (source of truth) + rune-templates (consumer) |
  | Curated + community-contributed template JSON | rune-templates |
  | API + gallery UI + seeder | rune |
  | AJV validation of bundle JSONs | rune-templates (CI) |

  ## How official vs community is decided

  The bundle JSON declares `"official": true` for Rune-team-curated templates; contributor PRs leave it false (the default) and land in the Community tab. PR review on `rune-templates` is the gate that prevents outside contributions from claiming official status.


ashmod:
This should ideally be merged after we populate rune-templates with an acceptable number of official templates in all categories.

Also, non-blocking (and probably will address in a followup), but we should study the possibility of using the validations from #631 here instead of having them duplicated. Currently, they live within the templates module so they are standalone and can be easily replaced when we need to.

PR #657, opened 2026-06-02
feat: role-based workflow filtering for workflows table

Lucifer3224:
**Summary**

Adds server support for filtering workflows by owner_id.
Introduces role-based table filtering behavior:

- **Admin**: filter by owner across all workflows.

- **Non-admin**: filter by All / Mine / Shared.

Keeps selected filter scope consistent after workflow actions.
Ensures table stats and runs reflect the active filter scope.


apps/web/src/components/workflows/table/WorkflowsTableFilters.tsx
```diff
@@ -78,15 +121,46 @@ export function WorkflowsTableFilters({
...
+            placeholder="Search workflows…"
+            value={query}
+            onChange={(event) => onQueryChange(event.target.value)}
+            className="pl-10"
+            aria-label="Search workflows"
+          />
+          <Search className="pointer-events-none absolute left-3 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground" />
+        </div>
+
+        {isAdmin && users.length > 1 && (
```
ashmod: The admin owner filter is hidden in some setups because `WorkflowsTableFilters` only renders it when users.length > 1, but `listUsersForSharing()` excludes the current user. With one admin and one viewer, the fetched list contains only the viewer, so `users.length === 1` and the admin sees no filtering control even though owner badges appear and `/workflows/?owner_id=` actually supports filtering by that owner.

Image

The admin filter should render when there is at least one selectable owner, and ideally should not use the sharing directory as its owner source because that endpoint is capped and excludes users by design.

apps/web/src/components/workflows/WorkflowsTable.tsx
```diff
@@ -192,16 +226,16 @@ export function WorkflowsTable() {
...
-    const draft = workflows.filter((workflow) => workflow.status === "draft").length;
-    const failed = workflows.filter(
+    const active = scopedWorkflows.filter((workflow) => workflow.status === "active").length;
+    const inactive = scopedWorkflows.filter((workflow) => workflow.status === "inactive").length;
+    const draft = scopedWorkflows.filter((workflow) => workflow.status === "draft").length;
+    const failed = scopedWorkflows.filter(
       (workflow) => lastRunByWorkflow.get(workflow.id)?.status === "failed",
     ).length;
-    const runs = executionItems.length;
+    const runs = scopedWorkflows.filter((workflow) => lastRunByWorkflow.has(workflow.id)).length;
```
ashmod, line 242: Is this intentional? I feel like the previous logic made more sense. Consider the use case where a user runs a workflow from that page twice, but the "Runs" counter does not change, their only indicator would be the "Last Run at x". I believe this could be a bit anticlimactic or misleading.

ashmod (commented):
Could we also add tests for this PR's new logic?

apps/web/src/components/workflows/WorkflowsTable.tsx
Lucifer3224: Yup, I changed it so the Runs stat matches what the Runs filter shows (workflows that have at least one execution), because showing e.g. 50 Runs but filtering down to 7 workflows felt inconsistent. Users who want total execution count can still see that in the Executions section.

I agree the current label can feel misleading when re-running a workflow from this page, since the counter won't tick up. Happy to rename the stat (e.g. "Executed" / "With runs") or revisit if you'd prefer keeping total run count here. Tell me your opinion.

apps/web/src/components/workflows/table/WorkflowsTableFilters.tsx
Lucifer3224: fixed

Lucifer3224:
> Could we also add tests for this PR's new logic?

Done

ashmod (approved):

PR #658, opened 2026-06-02
feat: add Jira integration with various functionalities

Spafic:
## Summary

Adds Jira integration with 10 operations: create/get/update/search issues, list/apply transitions, create/list comments, list projects, and get user.

## Changes

| Area | Files |
|------|-------|
| **Worker** | New `services/rune-worker/pkg/integrations/providers/jira/` with platform constants, helpers, and operation nodes. Registered in bootstrap loader. |
| **Frontend** | Jira integration entry in `catalog.ts` , registered in `nodeRegistry.ts`, SVG icon added. Updated `IntegrationInspector.tsx` and `types.ts`. |
| **Connector** | Extended `connector.go` to support API version selection. |
| **Docs** | `oauth-setup-steps.md` for Jira OAuth configuration, updated `README.md`. |

## Operations

- **Issues:** `issue_create`, `issue_get`, `issue_update`, `issue_search`
- **Transitions:** `issue_transition_list`, `issue_transition_apply`
- **Comments:** `comment_create`, `comment_list`
- **Projects:** `project_list`
- **Users:** `user_get`

Supports both Jira Cloud (API v3) and Jira Server/Data Center (API v2).


apps/web/src/features/canvas/components/inspectors/IntegrationInspector.tsx
```diff
@@ -122,7 +122,18 @@ export function IntegrationInspector({ node, updateData }: IntegrationInspectorP
...
+                typeof nextValue === "object" &&
+                nextValue !== null &&
+                !Array.isArray(nextValue) &&
+                Object.keys(nextValue).length === 0
+              ) {
+                updateArgument(field.name, undefined);
+              } else {
+                updateArgument(field.name, nextValue);
+              }
+            }}
```
ashmod: Three issues here:

1. This inline check is exactly `isJsonObject`, which is already defined at the top (L24) - you could simply use 
``` typescript
if (isJsonObject(nextValue) && Object.keys(nextValue).length === 0).
```
2. This fixes the symptom in one consumer, but the root cause is likely in `JsonField.tsx` since it initializes untouched fields to `"{}"` and `handleBlur` always emits - so just tabbing through a JSON field stores `{}`. Fixing it in JsonField (skip `onChange` when text is unchanged, or an `emptyAsUndefined` prop) covers all consumers, not just this one, if that's what you were aiming for.
3. Behavior is asymmetric: `{}` is dropped but `[]`, `null`, `""` are persisted. Also, after removal the value becomes `undefined`, and JsonField's logic re-renders `"{}"` - so the editor shows {} while the node stores nothing. UI state and persisted state silently disagree.

Also it's worth pointing out this changes behavior for every integration's JSON fields (Slack, Outlook, etc.), not just Jira, shouldn't be a problem but make sure no other integrations conflict with this in any way.

apps/web/src/features/canvas/integrations/catalog.ts
```diff
@@ -43,6 +49,34 @@ const BODY_CONTENT_TYPE_OPTIONS = [
...
+    type: "text",
+    placeholder: "d3c2f2e0-xxxx-xxxx-xxxx-acdeabcd1234",
+    help: "Optional. Leave empty for OAuth; it will be resolved from the site URL.",
+  },
+  {
+    name: "api_version",
+    label: "API version",
+    type: "select",
+    options: JIRA_API_VERSION_OPTIONS,
+  },
```
ashmod, line 89: I'm wondering why we have those as repeated fields for every node? Aren't they just inferred from the credential?

services/rune-worker/pkg/integrations/providers/jira/oauth-setup-steps.md
```diff
@@ -0,0 +1,99 @@
+# Jira OAuth 2.0 (3LO) Full Setup Steps
+
```
ashmod: Let's not commit this one, keep it draft somewhere locally or on a separate branch, and we'll add this to the website's docs (for all integrations) once they're up.

apps/web/src/features/canvas/integrations/catalog.ts
Spafic: These are currently per-node overrides because the worker resolves base_url, cloud_id, and api_version from args/params first, then credentials, and finally attempts cloud-id resolution for OAuth. That allows a single credential to target multiple Jira sites or override the cloud-id per node. If we want to infer-only from creds, I’m happy to revisit and remove the fields, but we’d need to confirm credential schema always includes site_url/cloud_id and that per-node overrides aren’t needed.

OAuth now don't have those fields

ashmod: Is there a jira-specific case where a user would use different values than the ones they entered in the credential? If not then just have them inferred, otherwise we don't have many options i guess

Spafic: The system works similarly to a sheet ID, offering two configuration methods: base_url or cloud_id. Users can choose whichever they find more convenient, as both function identically despite using different formats.

While a dedicated Jira credential could be created, it would unnecessarily complicate the process.

This approach mirrors how Google Sheet IDs are handled, but here we have 2 ways.

ashmod: Nice, let's make sure the inspector form is clear on that, have it display one or the other but not both.

Spafic: Do you want to disable one field if the other has a value? I added a note in the help text: "Optional. Leave empty for OAuth; it will be resolved from the site URL."

ashmod: Yes, the hint supports it but since only one field can be used at a time then only have one enabled. A simple user facing dropdown to select the type will do.

ashmod (approved):

Issue #441, opened 2026-03-15
Add pagination to workflows, executions, and credentials pages


ashmod:
## Feature Summary

Add server-backed pagination to the main list pages in the app, starting with:

- `/create/workflows`
- `/create/executions`
- `/create/credentials`

These pages should load one page of results at a time, with clear pagination controls and consistent list metadata from the API.

## Problem / Use Case

These screens currently fetch full datasets and then filter/render them on the client. That is manageable for small accounts, but it will not scale well as users accumulate more workflows, executions, and credentials.

This creates a few problems:
- slower page loads
- more client-side filtering work
- larger payloads than needed
- harder navigation through older items
- extra overhead on credentials, where row-level metadata is loaded after the main list

## Proposed Solution

Add a shared paginated list pattern across the API and frontend.

Backend:
- support pagination on `GET /workflows`, `GET /executions`, and `GET /credentials`
- return list metadata such as `page`, `page_size`, `total`, and `has_next`
- keep ordering deterministic so paging is stable
- move relevant filtering to the server where possible

Frontend:
- add pagination controls to the three pages above
- keep filters/search working with pagination
- reset to page 1 when filters change
- preserve good UX after create/delete actions

For executions specifically, make sure summary metrics still represent the full filtered result set, not just the current page.

## Additional Context

Suggested rollout order:
1. `/create/executions`
2. `/create/workflows`
3. `/create/credentials`

Follow-up candidates later:
- `/admin/users`
- `/create/templates`

## Acceptance Criteria

- workflows, executions, and credentials use server-backed pagination
- pagination controls are available and easy to use
- filtering/search works correctly with paged results
- API responses include pagination metadata
- execution metrics remain accurate after pagination is introduced

PR #662, opened 2026-06-03
feat: implement pagination for workflows list view

Lucifer3224:
**Summary**
- Integrates pagination and status filters into the workflows list backend API and frontend view.
- Adds pagination controls (range indicator, size selector, page navigators) to the existing table.
- Verified with new tests


services/api/src/workflow/router.py
```diff
@@ -41,13 +42,36 @@
...
-@router.get("/", response_model=ApiResponse[list[WorkflowListItem]])
+@router.get(
+    "/",
+    response_model=ApiResponse[
+        list[WorkflowListItem] | PaginatedData[WorkflowListItem]
+    ],
+)
 async def list_workflows(
+    page: int | None = None,
+    page_size: int | None = None,
```
ashmod: Could we add some sort of validation for those? Would also be worth adding a test for invalid ranges.

apps/web/src/components/workflows/WorkflowsTable.tsx
```diff
@@ -619,14 +698,28 @@ export function WorkflowsTable() {
...
+    }
+    return filtered.length;
+  }, [isExecutionFilter, pagination, workflows.length, filtered.length]);
+
+  const totalPages = useMemo(() => {
+    if (!isExecutionFilter) {
+      return pagination?.totalPages ?? 1;
+    }
+    return Math.ceil(totalCount / pageSize);
+  }, [isExecutionFilter, pagination, totalCount, pageSize]);
```
ashmod: Can we clamp `page` after deletes or when `totalPages` changes? Deleting the last item on the last page can leave `page > totalPages`, causing the table to show an empty page even though earlier pages still have data.

Image

ashmod (commented):
Looks solid. We also want to apply pagination to other pages as described in  #441. But this could be addressed in a follow-up PR.

Lucifer3224:
> Looks solid. We also want to apply pagination to other pages as described in #441. But this could be addressed in a follow-up PR.

I'm okay with either. let me know if you'd like me to implement them in this PR or open a new one, and I'll start immediately.

ashmod (dismissed):

ashmod:
> I'm okay with either. let me know if you'd like me to implement them in this PR or open a new one, and I'll start immediately.

Merging this one now, and let's have the other pages done in a separate PR

ashmod (approved):

PR #668, opened 2026-06-05
feat(Integration): Integrate with telegram

Spafic:
## Summary

Add Telegram Bot API integration — canvas nodes + worker execution for sending messages, photos, documents, fetching updates, and a Get Chat ID helper.

### Changes

- **Worker**: 5 Telegram nodes (send_message, send_photo, send_document, get_updates, get_chat_id) with credential-based auth, flat output structure, and `chat_id` type-safe conversion
- **Frontend**: Node registry, catalog entries with help text, credential filtering (token-only for Telegram), and a collapsible setup guide in the inspector


services/rune-worker/pkg/integrations/providers/telegram/send_message.go
```diff
@@ -0,0 +1,67 @@
...
+		body["disable_notification"] = true
+	}
+
+	return connector.Do(ctx, ec, connector.Spec{
+		Method:  "POST",
+		BaseURL: baseURL,
+		Path:    "/bot{token}/sendMessage",
+		PathArgs: map[string]string{
+			"token": token,
+		},
```
ashmod, line 60: This leaks the bot token into error msgs, we should redact it.

services/rune-worker/pkg/integrations/providers/telegram/get_chat_id.go
```diff
@@ -0,0 +1,133 @@
...
+
+	raw, err := connector.Do(ctx, ec, connector.Spec{
+		Method:  "POST",
+		BaseURL: baseURL,
+		Path:    "/bot{token}/getUpdates",
+		PathArgs: map[string]string{
+			"token": token,
+		},
+		Body: map[string]any{
+			"limit": 1,
```
ashmod, line 34: Telegram returns pending updates oldest-first, so `limit: 1` selects the oldest update rather than the “latest” one advertised by this node. With messages from multiple chats, this can return an unrelated chat ID. Please fetch enough updates to select the highest `update_id`, or otherwise explicitly request the newest update.

apps/web/src/features/canvas/components/inspectors/IntegrationInspector.tsx
```diff
@@ -183,13 +183,47 @@ export function IntegrationInspector({ node, updateData }: IntegrationInspectorP
...
+              <strong>3. Find your Chat ID</strong>
+              <br />
+              Message your bot on Telegram first (any text), then use the{" "}
+              <span className="text-foreground/70">Get Chat ID</span> node to auto-detect it. Copy
+              the number into the <span className="font-mono text-foreground/70">chat_id</span>{" "}
+              field of other Telegram nodes.
+            </p>
+          </div>
+        </details>
+      )}
```
ashmod, line 236: Add a TODO to remove this section and instead link it to the docs page when ready.

services/rune-worker/pkg/integrations/internal/connector/connector.go
```diff
@@ -30,7 +30,7 @@ type Error struct {
 }
 
 func (e *Error) Error() string {
-	return fmt.Sprintf("integration request failed: status=%d url=%s", e.Status, e.URL)
+	return fmt.Sprintf("integration request failed: status=%d url=%s body=%v", e.Status, e.URL, e.Body)
```
ashmod: The added `body=%v` will now affect all connectors not just Telegram's. So this may leak sensitive info for other integrations as well, if that's the case and we're only adding this for debugging, then we should remove this before merging.

apps/web/src/features/canvas/integrations/catalog.ts
```diff
@@ -628,6 +639,166 @@ export const INTEGRATION_TOOLS = [
...
+        placeholder: "123456",
+        help: "Use the Get Chat ID node output or paste the numeric ID directly.",
+      },
+      { name: "text", label: "Message", type: "textarea", required: true },
+      {
+        name: "parse_mode",
+        label: "Parse mode",
+        type: "select",
+        options: PARSE_MODE_OPTIONS,
+        help: "HTML and MarkdownV2 let you use bold, italic, links, etc. None sends plain text.",
```
ashmod: The hint says "None sends plain text" but there is actually no "None" option, the user is forced to an HTML/MDv2 dropdown.

Spafic: None means no `<b>` tag in html or `_ _`  in md like just a string

ashmod: The help text is suggesting the user can select "none" of the options described earlier. But they can't, the dropdown forces a selection from the two. Either remove that part of the hint, or add a way to deselect (e.g. a "Choose ..." on top of the dropdown list.)

ashmod (approved):

PR #673, opened 2026-06-12
feat: implement pagination for credentials and executions views

### LINKED ISSUE #441: Add pagination to workflows, executions, and credentials pages

Lucifer3224:
Summary

* Integrates pagination, search, and type/status filters into the credentials and executions list backend APIs and frontend views.
* Adds pagination controls (range indicator, size selector, page navigators) to the existing credentials and executions tables.
* Verified with new tests

closes #441


ashmod (approved):
LGTM , but we need a follow-up for the "Workflows" view in the Executions table.

Row-based pagination doesn't fit the grouped view: a page can collapse into a single workflow, groups get split across pages, and the header badge count only reflects the rows on the current page rather than the workflow's actual executions, which is misleading.

Grouped mode should instead paginate **per** workflow: page over workflows, each group preloading its N most recent executions plus real aggregates for the header. That probably needs a `group_by=workflow` variant of the endpoint with its own response shape, so we should target it in a follow-up issue rather than growing this PR.

